PR #29, opened 2019-12-04
Prepare API client

andrzejewsky:
# How it works

Commerce tools has a graphql API. To do graphql calls we used apollo library. The apollo configuration is pretty straightforward here, except one thing which is access token that is required by commerce tools platform. Every single api call requires access token that must be fetched beforehand. To cover that case we used customised apollo link, that asks for the token, and stores it in the `localStorage` (to avoid multiple requests all time).

## setup commerce tools api

To setup commerce tools api you have to use `serup` function that accepts two params:
- `config` - contains required data to the comemrcetools api such as clientId, projectKey, hosts etc.
- `customOptions` - additional options that you want to add to the apollo client configuration

```javascript
import { setup } from '@vue-storefront/commercetools-api'

setup({
  config: {
    uri: 'https://api.example.com/graphql',
    authHost: 'https://auth.example.com',
    projectKey: '....',
    clientId: '....',
    clientSecret: '....',
    scopes: ['manage_products:project'],
  }
})
```

## write new api calls

Place a new function inside `src/api` directory, import the apolloClient (that has just configured by `setup`) and call the query.

Example:

```javascript
import { apolloClient } from './../index'
import gql from 'graphql-tag'

const getProduct = async () => {
  return await apolloClient.query({
    query: gql`
      {
        product(id: "c8d1fb5e-c3ae-4428-b895-66a17486e6a9") {
          id
          key
          version
        }
      }
    `,
  })
}

export default getProduct
```

## testing

To test particular api call, create a new test file in `tests/api` and use your function. 

example:

```javascript
import getProduct from './../../src/api/getProduct';

describe('test getProduct', () => {
  it('fetches product', async () => {
    const {
      data: { product },
    } = await getProduct();
    expect(product).toEqual({
      id: 'test id',
      key: 'example key',
      version: 2,
      __typename: 'Product',
    });
  });
});
```

## mocking

In most cases, when you have to test your new endpoint, you have to write response mock. To do this, you have to write your own resolvers. All of them should be placed in `tests/resolvers` directory. 

example: 
```javascript
const resolvers = {
  Query: {
    product: () => {
      return { id: 'test id', key: 'example key', version: 2 }
    }
  }
}

export default resolvers
```

### How it works?

In the test environment, we create a different apollo client. Instead of HttpLink we have SchemaLink and instead of network calls, we have our own resolvers. This give us possibility to test the query and mock example responses that are expected from the API.


packages/commercetools/api-client/package.json
```diff
@@ -2,15 +2,34 @@
...
+    "apollo-client": "^2.6.4",
+    "apollo-link-context": "^1.0.19",
+    "apollo-link-http": "^1.5.16",
+    "cross-env": "^6.0.3",
+    "graphql": "^14.5.8",
+    "graphql-tag": "^2.10.1",
+    "isomorphic-fetch": "^2.2.1"
   },
   "devDependencies": {
+    "@types/jest": "^24.0.23",
```
filrak: i think we can pull most of these dependencies to root package as they are purely dev-related.

packages/commercetools/api-client/src/api/getProduct.ts
```diff
@@ -0,0 +1,18 @@
+import { apolloClient } from './../index'
+import gql from 'graphql-tag'
+
+const getProduct = async () => {
```
filrak, line 5: missing params and types

packages/commercetools/api-client/src/api/getProduct.ts
```diff
@@ -0,0 +1,18 @@
...
+  return await apolloClient.query({
+    query: gql`
+      {
+        product(id: "c8d1fb5e-c3ae-4428-b895-66a17486e6a9") {
```
filrak, line 9: do we really need to have it hardcoded? shouldn't we use tests for it?

filrak (changes requested):

filrak:
how `customOptions` work? What exactly we can pass here?

packages/commercetools/api-client/src/api/getProduct.ts
andrzejewsky: I'll do it later, this is just example, in the next tasks we have API implementation

packages/commercetools/api-client/src/api/getProduct.ts
andrzejewsky: same as above

andrzejewsky:
@filrak so customOptions are the options that you can pass to ApolloClient (description in PR). This options are represent by type `ApolloClientOptions`. Basically you can set there your own configuration such as custom link, cache, even resolvers, type definition etc.(more: https://www.apollographql.com/docs/react/api/apollo-client/#ApolloClient)

I left possibility to add these options because sometimes they are useful, eg. to set custom cache policy or something like that.

PR #45, opened 2019-12-10
Category composable and page basic setup

filrak:
Scope:
- [x] dummy composable
- [x] use dummy composable in page
- [x] type core interfaces with Refs
- [x] create generic category page with multi-level category URL structure `/c/slug/dlug2...`
- [ ] tests

Tests to be done in next PR once i have non-dummy data

### DIFF
--- a/packages/commercetools/composables/src/useCategory/index.ts
+++ b/packages/commercetools/composables/src/useCategory/index.ts
@@ -1,0 +1,22 @@
+import { UseCategory } from '@vue-storefront/interfaces'
+import { ref } from '@vue/composition-api'
+
+export default function useCategory (): UseCategory<any, any, any, any, any> {
+  const category = ref(null)
+  const search = () => { console.log('useCategory:search') }
+  const appliedFilters = ref(null)
+  const applyFilter = () => { () => { console.log('useCategory:applyFilter') } }
+  const clearFilters = () => { () => { console.log('useCategory:clearFilters') } }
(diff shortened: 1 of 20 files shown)


packages/commercetools/composables/src/useCategory/index.ts
```diff
@@ -0,0 +1,27 @@
+import { UseCategory } from '@vue-storefront/interfaces'
+import { ref } from '@vue/composition-api'
+import { getProduct } from '@vue-storefront/commercetools-api'
+
+export default function useCategory (): UseCategory<any, any, any, any, any> {
```
andrzejewsky, line 5: why useCategory? shouldn't be useProduct? we want to load and show products not categories yet?

packages/commercetools/composables/src/useCategory/index.ts
```diff
@@ -0,0 +1,27 @@
...
+  const category = ref({})
+  const search = () => { console.log('useCategory:search') }
+  const appliedFilters = ref(null)
+  const applyFilter = () => { () => { console.log('useCategory:applyFilter') } }
+  const clearFilters = () => { () => { console.log('useCategory:clearFilters') } }
+  const loading = ref(false)
+  const error = ref(null)
+
+  getProduct().then(res => {
```
andrzejewsky, line 14: why composable function useCategory fetches products?

andrzejewsky (changes requested):

PR #47, opened 2019-12-10
Implement getProduct

andrzejewsky:


packages/commercetools/theme/pages/index.vue
```diff
@@ -155,12 +155,23 @@ import {
...
+  return encodeURIComponent(text).replace(/[!'()~*]/g, (c) => {
+    return '%' + c.charCodeAt(0).toString(16);
+  });
+}
 
 export default {
   name: "Home",
   transition: 'fade',
   data() {
+    getProduct({ variables: {
```
andrzejewsky: @filrak example how to use it, I'll implement a generic helpers (eg. `getWhereByCatId`, `getWhereByCatSlug` etc.) to generate `where` param

packages/commercetools/api-client/src/api/getProduct/defaultQuery.ts
```diff
@@ -0,0 +1,60 @@
+import gql from 'graphql-tag'
```
filrak, line 1: can't we just use gql files?

packages/commercetools/api-client/src/helpers/createCommerceToolsLink/getAccessToken.ts
```diff
@@ -1,7 +1,7 @@
 import SdkAuth from '@commercetools/sdk-auth'
-import { ApiConfig } from '../../types/setup'
-import { storeToken, getToken } from './tokenCache'
-import { Token } from '../../types/setup'
+import { ApiConfig } from 'api-client/src/types/setup'
```
filrak, line 2: maybe it's better to just use '@' alias for root as common convention

packages/commercetools/api-client/src/helpers/i18n/index.ts
```diff
@@ -0,0 +1,5 @@
+const getCurrency = () => 'USD'
```
filrak: Imho it's better to keep it from config passed into `setup` then in the app you can set reference to currency and locale fields. It's also better to keep all configuration in one place

filrak (changes requested):

PR #60, opened 2019-12-13
api(category) getCategory implementation

andrzejewsky:


packages/commercetools/api-client/src/api/getCategory/defaultQuery.ts
```diff
@@ -0,0 +1,27 @@
+import gql from 'graphql-tag'
```
filrak: Accoridng to our discussion - we're missing products here

filrak (approved):

PR #63, opened 2019-12-17
ui(ProductPage) add dummy product page route and layout

### LINKED ISSUE #62: I want to display hardcoded Product Page layout So then I can start working on its logic
**Context**
At this stage we do not have Product Page implemented at all. There will be some more complex logic. Right now we only want to display standard Storefront UI page. In the next steps, we will add the possibility to redirect from Category Page to Product Page.

**Acceptance Criteria**
* Product Page from Storefront UI is used https://storybook.storefrontui.io/?path=/story/pages-product--default
* I can see some hardcoded dummy data in the page.
* I can see hardcoded image gallery.
* Everything is properly aligned just like in Storefront UI storybook.
* I can enter the Product Page by using direct URL in a browser.
* No additional logic that calls commercetools API is implemented.

EugeneHlushko:
- add route `/p/slug` to handle product pages alongside to `/l/slug` of categories
- add page layout with dummy content

To get to the page just visit any `/p/fsa` url

Closes #62

### DIFF
--- a/packages/commercetools/theme/nuxt.config.js
+++ b/packages/commercetools/theme/nuxt.config.js
@@ -18,12 +18,12 @@ export default {
       // @todo move to common middleware in core module and cover use case where we don't have slug at all
       // aka URL dispatcher
       routes.push({
-        name: 'customCategory',
+        name: 'category',
         path: '/l/:slug_1/:slug_2?/:slug_3?/:slug_4?/:slug_5?',
         component: resolve(__dirname, 'pages/Category.vue')
       })
       routes.push({
-        name: 'customProduct',
+        name: 'product',
         path: '/p/:slug_1/',
         component: resolve(__dirname, 'pages/Product.vue')
       })
--- a/packages/commercetools/theme/nuxt.config.js
+++ b/packages/commercetools/theme/nuxt.config.js
@@ -18,9 +18,14 @@ export default {
       // @todo move to common middleware in core module and cover use case where we don't have slug at all
       // aka URL dispatcher
       routes.push({
-        name: 'custom',
+        name: 'customCategory',
         path: '/l/:slug_1/:slug_2?/:slug_3?/:slug_4?/:slug_5?',
         component: resolve(__dirname, 'pages/Category.vue')
+      })
+      routes.push({
+        name: 'customProduct',
+        path: '/p/:slug_1/',
+        component: resolve(__dirname, 'pages/Product.vue')
       })
     }
   },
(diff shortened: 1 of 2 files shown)


packages/commercetools/theme/pages/Product.vue
```diff
@@ -0,0 +1,738 @@
...
+  },
+  methods: {
+    toggleWishlist(index) {
+      this.products[index].isOnWishlist = !this.products[index].isOnWishlist;
+    }
+  }
+};
+</script>
+<style lang="scss" scoped>
+@import "~@storefront-ui/vue/styles";
```
filrak: we have those styles already imported in a root.

filrak (changes requested):

packages/commercetools/theme/nuxt.config.js
```diff
@@ -18,10 +18,15 @@ export default {
       // @todo move to common middleware in core module and cover use case where we don't have slug at all
       // aka URL dispatcher
       routes.push({
-        name: 'custom',
+        name: 'customCategory',
```
filrak: I think we can just rename those to `category` and `product`, wdyt @EugeneHlushko ?

EugeneHlushko: yup makes sense to me, will push an update

packages/commercetools/theme/pages/Product.vue
EugeneHlushko: Still have to import variables tho

EugeneHlushko:
@filrak just updated

packages/commercetools/theme/pages/Product.vue
filrak: yeah but you can import them directly from `@storefront-ui/src/css/variables` i'm not sure if importing whole stylesheet in scoped component won't end up in duplicating it, you can either check iif styles are duplicated or just change import to point to vars directory. Tbh I'm curious ;)

EugeneHlushko: yup thats what i did yesterday after your comment: `@import "~@storefront-ui/shared/styles/variables";`

This (nuxt) does use rollup so i am not sure as well, i know the answer for webpack though it depends on config

filrak: I'm 100% sure nuxt is using webpack. I am precisely interested in `scoped` attribute of CSS part - it scopes alll it's styles with prefixes so there is a chance it will also duplicate the imported ones

EugeneHlushko: right it does, i was pretty sure to read something like that on twitter :D

EugeneHlushko: regarding this PR, is there anything you want to improve or can it be merged?

PR #72, opened 2019-12-21
helpers(product) implement product helpers

andrzejewsky:


packages/commercetools/helpers/src/index.ts
```diff
@@ -1 +1,15 @@
-console.log('Hello commercetools helpers')
\ No newline at end of file
+import { UiMediaGalleryItem, UiProductName, UiProductPrice } from '@vue-storefront/interfaces'
+import { ProductVariant, Image, ProductData } from './types/GraphQL'
+
+export const getName = (product: ProductData): UiProductName => product.name
```
filrak: You're missing `Product` in all helper names -> eg `getProductName` etc

packages/commercetools/helpers/src/index.ts
```diff
@@ -1 +1,15 @@
...
+
+export const getSlug = (product: ProductData): UiProductName => product.slug
```
filrak: How about this `Agnostic` instead of UI eg. `AgnosticProductName` ? 
Also for simple strings/numbers maybe we should keep it as string/number and add those types only for objects to not confuse people? String is string. I don't see a value in renaming it 
@patzick

packages/commercetools/helpers/src/index.ts
```diff
@@ -1 +1,15 @@
-console.log('Hello commercetools helpers')
\ No newline at end of file
+import { UiMediaGalleryItem, UiProductName, UiProductPrice } from '@vue-storefront/interfaces'
```
filrak: You're missing `getVariant` which is pretty important since rest of the operations is on variants

packages/core/interfaces/index.ts
```diff
@@ -108,4 +108,16 @@ export interface UseContent<CONTENT, SEARCH> {
...
   error: any
-}
\ No newline at end of file
+}
+
+export interface UiMediaGalleryItemUrl {
+  url: string
+}
+export interface UiMediaGalleryItem {
+  small: UiMediaGalleryItemUrl
```
filrak: Wouldn't it be misleading in autosuggestions? User won't see if thats a string or something else. While typing response makes sense to em renaming strings to other properties desn't and is confusing
Again @patzick  for validation

filrak (changes requested):

packages/core/interfaces/index.ts
filrak: Here is an example of confusion - user even doesnt know that its a string 
image

packages/commercetools/api-client/src/api/getProduct/index.ts
```diff
@@ -5,22 +5,23 @@ import { ProductSearch } from './../../types/Api'
 import { ProductQueryResult } from './../../types/GraphQL'
 import defaultQuery from './defaultQuery'
 import { buildProductWhere } from './../../helpers/search'
+import { applyInternals } from './../../helpers/internals'
```
filrak: maybe we could name it better ;) There might be more helper properties for different entities. How about something more descriptive like `mixParentDataToVariants` or sth like this?

packages/commercetools/composables/src/helpers/internals/enhanceCategory.ts
```diff
@@ -0,0 +1,30 @@
+import { ApolloQueryResult } from 'apollo-client'
```
filrak, line 1: maybe mixins is a better name instead of "internals"?

packages/commercetools/composables/src/helpers/internals/enhanceCategory.ts
```diff
@@ -0,0 +1,30 @@
...
+}
+
+interface ProductData {
+  products: ProductQueryResult
+}
+
+type ProductResponse = ApolloQueryResult<ProductData>
+type CategoryResponse = ApolloQueryResult<CategoryData>
+
+const applyProducts = (rawProducts: ProductVariant[]) => (category: Category) => {
```
filrak: mixProducts? just an idea for convention, maybe stupid

packages/commercetools/composables/src/useCategory/index.ts
```diff
@@ -1,12 +1,22 @@
...
 import { getCategory, getProduct } from '@vue-storefront/commercetools-api'
+import { enhanceProduct, enhanceCategory } from './../helpers/internals'
 
 interface UseCategorySearchParams {
-  /** Category ID  */
   slug?: string
 }
 
+const loadCategories = async (params: UseCategorySearchParams) => {
+  const categoryResponse = await getCategory({ slug: params.slug })
```
filrak: we should be able to search through all available params of `getCategory`

packages/commercetools/helpers/src/index.ts
```diff
@@ -1 +1,29 @@
...
+
+export const getProductName = (product: ProductVariant): string => (product as any)._name
+
+export const getProductSlug = (product: ProductVariant): string => (product as any)._slug
+
+export const getProductPrice = (product: ProductVariant): number => product.price.value.centAmount
+
+export const getProductGallery = (product: ProductVariant): UiMediaGalleryItem[] =>
+  product.images.map((image: Image) => ({
+    small: '',
```
filrak: if we dont have alternative sizes shouldnt we just provide same img everywhere? If someone will use small ones on one platform and change the platform to CT they will be blank. If we fill all with same image they will still be visible.

packages/commercetools/helpers/src/index.ts
```diff
@@ -1 +1,29 @@
-console.log('Hello commercetools helpers')
\ No newline at end of file
+import { UiMediaGalleryItem } from '@vue-storefront/interfaces'
```
filrak, line 1: Bjutiful <3

filrak (changes requested):

packages/commercetools/composables/src/helpers/internals/enhanceCategory.ts
andrzejewsky: I was thinking about the name for these "private" fields, and I think `internals` is the best one as it describes the purpose - they are fields or logic that we will use **internally**  it could be the mixins but it might be confusing with the vue mixins. I'll keep it as it is now (intenals) but maybe we will come up with a different name

PR #74, opened 2019-12-23
ui(SideCart) add side cart component and composable

EugeneHlushko:
- Side cart component
- ui-state
- header cart icon triggers cart on side
- cart is closable

### DIFF
--- a/packages/commercetools/composables/src/useCart/index.ts
+++ b/packages/commercetools/composables/src/useCart/index.ts
@@ -1,0 +1,63 @@
+import { UseCart } from '@vue-storefront/interfaces'
+import { ref } from '@vue/composition-api'
+
+export default function useCart(): UseCart<any, any, any, any, any, any, any> {
+  const cart = ref({ products: [
+    {
+      title: "Cream Beach Bag",
+      id: "CBB1",
+      image: "/homepage/productA.jpg",
+      price: { regular: "50.00" },
+      configuration: [
+        { name: "Size", value: "XS" },
+        { name: "Color", value: "White" }
+      ],
+      qty: "1",
+      stock: 44
+    },
+    {
+      title: "Cream Beach Bag",
+      id: "CBB2",
+      image: "/homepage/productB.jpg",
+      price: { regular: "50.00", special: "20.05" },
+      configuration: [
+        { name: "Size", value: "XS" },
+        { name: "Color", value: "White" }
+      ],
+      qty: "2",
+      stock: 10
+    },
+    {
+      title: "Cream Beach Bag",
+      id: "CBB3",
+      image: "/homepage/productC.jpg",
+      price: { regular: "50.00", special: "20.50" },
+      configuration: [
+        { name: "Size", value: "XS" },
+        { name: "Color", value: "White" }
+      ],
+      qty: "1",
+      stock: 20
+    }
+  ]})
+  const addToCart = () => { () => { console.log('useCart:addToCart') } }
+  const removeFromCart = () => { () => { console.log('useCart:removeFromCart') } }
+  const clearCart = () => { () => { console.log('useCart:clearCart') } }
+  const coupon = ref({})
+  const applyCoupon = () => { () => { console.log('useCart:applyCoupon') } }
+  const removeCoupon = () => { () => { console.log('useCart:removeCoupon') } }
+  const loading = false
+  const error = ref(null)
+
+  return {
+    cart,
+    addToCart,
+    removeFromCart,
+    clearCart,
+    coupon,
+    applyCoupon,
+    removeCoupon,
+    loading,
+    error
+  }
+}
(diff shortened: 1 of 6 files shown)


packages/commercetools/composables/src/useCart/index.ts
```diff
@@ -0,0 +1,63 @@
+import { UseCart } from '@vue-storefront/interfaces'
+import { ref } from '@vue/composition-api'
+
+export default function useCart(): UseCart<any, any, any, any, any, any, any> {
+  const cart = ref({ products: [
```
EugeneHlushko, line 5: This is dummy data hardcoded

packages/commercetools/theme/components/SideCart.vue
```diff
@@ -0,0 +1,212 @@
+<template>
+  <div id="cart">
+    <SfSidebar
+      :visible="isCartSidebarOpen"
```
EugeneHlushko, line 4: How are we planning to handle this outside of component, say i added to cart from another component somewhere down the layout tree, how do i dispatch that event and make sure this hears it @filrak ? In React for example, we could use state on the top level component and hook into it from both, or use redux. Whats the vue way?

packages/commercetools/theme/components/AppHeader.vue
```diff
@@ -18,7 +18,7 @@
       </nuxt-link>
       <nuxt-link to="/l/cat">
         <SfHeaderNavigationItem>
-          KIDS
+          {{ isCartSidebarOpen }}
```
EugeneHlushko: will cleanup

packages/commercetools/theme/components/AppHeader.vue
```diff
@@ -27,12 +27,23 @@
...
 import { SfHeader, SfImage } from '@storefront-ui/vue'
+import { isCartSidebarOpen } from '../assets/ui-state'
 
 export default {
   components: {
     SfHeader,
     SfImage
   },
+  methods: {
+    showCartSidebar: () => {
```
EugeneHlushko, line 38: point of interest

packages/commercetools/theme/components/AppHeader.vue
```diff
@@ -27,12 +27,23 @@
...
+      isCartSidebarOpen.value = true
+    }
+  },
+  setup() {
+    return {
+      isCartSidebarOpen: isCartSidebarOpen.value
```
EugeneHlushko: point of interest

packages/commercetools/theme/assets/ui-state.js
```diff
@@ -0,0 +1,17 @@
+import Vue from 'vue'
+import VueCompositionApi, { reactive, ref, computed } from '@vue/composition-api'
+
+Vue.use(VueCompositionApi);
```
filrak, line 4: It's already registered in nuxt moduł so you can just use it as it is

packages/commercetools/theme/assets/ui-state.js
```diff
@@ -0,0 +1,17 @@
...
+
+const UiState = reactive({
+  isCartSidebarOpen: false
+})
+
+const isCartSidebarOpen = computed({
+  get: () => UiState.isCartSidebarOpen,
+  set: val => { UiState.isCartSidebarOpen = val }
```
filrak: I think it's safer to have a toggle method. Then user can override this variable, just change its value

packages/commercetools/theme/assets/ui-state.js
```diff
@@ -0,0 +1,17 @@
+import Vue from 'vue'
+import VueCompositionApi, { reactive, ref, computed } from '@vue/composition-api'
+
+Vue.use(VueCompositionApi);
+
+const UiState = reactive({
+  isCartSidebarOpen: false
+})
+
+const isCartSidebarOpen = computed({
```
filrak: How about grouping it all in 'ui' object for autocompletr purposes?

packages/commercetools/theme/components/AppHeader.vue
```diff
@@ -27,12 +27,18 @@
...
+import { isCartSidebarOpen } from '../assets/ui-state'
 
 export default {
   components: {
     SfHeader,
     SfImage
   },
+  methods: {
+    showCartSidebar: () => {
+      isCartSidebarOpen.value = true
```
filrak: You can directly pass the 'ui' object to 'setup' method

filrak (changes requested):

packages/commercetools/theme/assets/ui-state.js
EugeneHlushko: It doesnt work without:

Screen Shot 2019-12-24 at 9 48 53 AM

EugeneHlushko:
Hi @filrak 
Just updated

packages/commercetools/theme/assets/ui-state.js
filrak: This is really weird since the plugin is registered already (setup function requires it to run). I think it needs some further investigation

PR #82, opened 2019-12-30
composables(useProduct): useProduct implementation

andrzejewsky:


packages/commercetools/api-client/src/helpers/search/index.ts
```diff
@@ -7,6 +7,10 @@ const buildProductWhere = (search: ProductSearch) => {
     return `masterData(current(categories(id in ("${catIds}"))))`
   }
 
+  if (search && search.slug) {
```
defudef, line 10: What about using ternary operator?

```JS
return search && search.slug
  ? `masterData(current(slug(${locale}="${search.slug}")))`
  : ''
```

packages/commercetools/composables/src/useProduct/index.ts
```diff
@@ -1,16 +1,38 @@
...
+  slug?: string
+  catIds?: string[]
+  skus?: string[]
+}
+
+type Search = (params: UseProductSearchParams) => void
+
+const loadProductVariants = async (params: UseProductSearchParams): Promise<ProductVariant[]> => {
+  const productResponse = await getProduct(params)
+  const enhancedProductResponse = enhanceProduct(productResponse)
```
defudef, line 17: Isn't better to define a type or interface? What do you think?

```TS
interface EnhanceProductResponse {
  data: {
    _variants: any
  }
}
```

packages/commercetools/helpers/src/index.ts
```diff
@@ -1,19 +1,23 @@
...
 import { ProductVariant, Image, Category } from './types/GraphQL'
 
-export const getProductName = (product: ProductVariant): string => (product as any)._name
+export const getProductName = (product: ProductVariant): string => product ? (product as any)._name : ''
 
-export const getProductSlug = (product: ProductVariant): string => (product as any)._slug
+export const getProductSlug = (product: ProductVariant): string => product ? (product as any)._slug : ''
 
-export const getProductPrice = (product: ProductVariant): number => product.price.value.centAmount
+export const getProductPrice = (product: ProductVariant): number => product ? product.price.value.centAmount : null
```
defudef: wrong return type. It's now `number | null`

packages/commercetools/helpers/src/index.ts
```diff
@@ -1,19 +1,23 @@
...
 
-export const getProductGallery = (product: ProductVariant): UiMediaGalleryItem[] =>
-  product.images.map((image: Image) => ({
+export const getProductGallery = (product: ProductVariant): UiMediaGalleryItem[] => {
+  if (!product) {
+    return []
+  }
+
+  return product.images.map((image: Image) => ({
```
defudef: Just an idea, but I'm not sure you'll like it:
```TS
export const getProductGallery = (product: ProductVariant): UiMediaGalleryItem[] => (product ? product.images : [])
  .map((image: Image)) => ({ ....other stuff })
```

packages/commercetools/helpers/src/index.ts
```diff
@@ -27,3 +31,15 @@ export const getCategoryProducts = (category: Category, options: any = {}): Prod
 
   return _products
 }
+
+export const getProductVariants = (products: ProductVariant[], options: any = {}): ProductVariant | ProductVariant[]  => {
```
defudef, line 32: I've seen a couple of times that option master true. Maybe it's worth to add an interface for it?

defudef (approved):

packages/commercetools/composables/src/useProduct/index.ts
andrzejewsky: underscored fields are `private` for us. we decided to use them to avoid custom typing for eCommerce platforms. Instead of modifying response (that are already typed) we are adding new fields (underscored)

packages/commercetools/helpers/src/index.ts
andrzejewsky: makes sense

packages/commercetools/helpers/src/index.ts
andrzejewsky: that's right, but there are different logic, for finding master one

PR #90, opened 2020-01-07
ui(cart) Issue/57/new

filrak:
updated https://github.com/DivanteLtd/next/pull/74/f
- renamed `setIsCartSidebarOpen()` to `toggleCartSidebar()` slight change: r
- made `isCartSidebarOpen` local tstae variable directly immutable, you can do this only in expected way via mutator (`togglecartSidebar`)

Composition API Plugin registration is required, no idea why. Needs to stay till Vue 3 ;x


packages/commercetools/theme/assets/ui-state.ts
```diff
@@ -0,0 +1,14 @@
+import Vue from 'vue'
+import VueCompositionAPI, { ref, computed } from '@vue/composition-api'
+
+// We need to register it again because of Vue instance instantiation issues
+Vue.use(VueCompositionAPI)
+
+const state = ref({
```
andrzejewsky: `reactive` i think

packages/commercetools/theme/components/CartSidebar.vue
```diff
@@ -0,0 +1,215 @@
...
+    removeHandler(product) {
+      const products = [...this.products];
+      this.products = products.filter(element => element.id !== product.id);
+    }
+  }
+};
+</script>
+<style lang="scss" scoped>
+@import "~@storefront-ui/shared/styles/variables";
+@mixin for-desktop {
```
andrzejewsky, line 136: I think would be nice to create another file to store all generic mixins

andrzejewsky (approved):

EugeneHlushko (approved):
LGTM, we can have a separate task for mixins

PR #93, opened 2020-01-07
Cart api

andrzejewsky:


packages/commercetools/api-client/src/api/createCart/defaultMutation.ts
```diff
@@ -0,0 +1,47 @@
...
+  }
+
+  mutation createCart($draft: CartDraft!, $locale: Locale!, $storeKey: KeyReferenceInput) {
+    cart: createCart(draft: $draft, storeKey: $storeKey) {
+      id
+      customerId
+      customerEmail
+      lineItems {
+        id
+        productId
```
filrak, line 24: can we add attributes here as well?

filrak: use case -> configuring products from cart

packages/commercetools/api-client/src/index.ts
```diff
@@ -17,8 +22,21 @@ const setup = <TCacheShape>(setupConfig?: SetupConfig<TCacheShape>): ApolloClien
   })
   locale = setupConfig.locale
   currency = setupConfig.currency
+  productAttributesIncluded = setupConfig.productAttributesIncluded
```
filrak: 1. It needs to be global, not scoped to cart
2. By default there shouldn't be any limit, field should be optional

packages/commercetools/api-client/src/types/Api.ts
```diff
@@ -1,24 +1,34 @@
-interface CustomQuery {
+import { ApolloQueryResult } from 'apollo-client'
+import { FetchResult } from 'apollo-link'
+import { Cart } from './GraphQL'
+
+export interface CustomQuery {
```
filrak, line 5: i still think custom queries at this point are bad idea, they are breaking all helper-related functionalities and (possibly) composables. If we want to keep it at least log a warining that using custom query can break the functionality

packages/commercetools/api-client/tests/setup.ts
```diff
@@ -3,4 +3,9 @@ import { setup } from './../src/index'
 jest.mock('./../src/helpers/createCommerceToolsLink')
 jest.mock('apollo-client')
 
-setup({ api: null, locale: 'en', currency: 'USD' })
+setup({
+  api: null,
+  locale: 'en',
+  currency: 'USD',
+  productAttributesIncluded: ['color', 'size']
```
filrak: do we see any other potential product config options? if so maybe we should put it under `product.attributesIncluded`

packages/commercetools/composables/src/useCart/currentCart.ts
```diff
@@ -0,0 +1,46 @@
+import { getCart, createCart, getStorage } from '@vue-storefront/commercetools-api'
```
filrak, line 1: shouldnt getStorage be internal helper?

packages/commercetools/composables/src/useCart/actions.ts
```diff
@@ -0,0 +1,24 @@
+import { ProductVariant } from "@vue-storefront/commercetools-api/lib/src/types/GraphQL"
+import { UiCartProduct } from '@vue-storefront/interfaces'
+
```
filrak: please add that 'action| is a name from ct ideally with link, for people new to this platform that will want to maintain this file naming could be tricky

packages/commercetools/composables/src/useCart/process.ts
```diff
@@ -0,0 +1,36 @@
+import { updateCart } from '@vue-storefront/commercetools-api'
```
filrak: As we agreed, lets move those processors to API Client

packages/commercetools/composables/tests/mountComposable.ts
```diff
@@ -0,0 +1,15 @@
+import { shallowMount } from '@vue/test-utils'
```
filrak: if its not a test maybe `_` at the beginning of file to indicate its a helper?

packages/core/interfaces/index.ts
```diff
@@ -34,15 +34,17 @@ export interface UseCart
   COUPON,
   APPLY_COUPON,
   REMOVE_COUPON,
+  UPDATE_QUANTITY,
 > {
   cart: CART;
   addToCart: ADD_TO_CART;
   removeFromCart: REMOVE_FROM_CART;
   clearCart: CLEAR_CART;
+  updateQuantity: UPDATE_QUANTITY;
```
filrak: as discussed - just extend composable interface

filrak (changes requested):

packages/commercetools/composables/src/useCart/currentCart.ts
andrzejewsky: I need this in composables and api-client as well

packages/commercetools/api-client/src/types/Api.ts
andrzejewsky: this is the old type that was created for category and products api - you can put custom queries there. Of course, before the final release, we have to give a possibility to pass a custom query for all API calls, but of course, it must be handled in the helpers (warning/error)

packages/commercetools/api-client/src/api/createCart/defaultMutation.ts
andrzejewsky: they are in the variant field -> attributesRaw

PR #130, opened 2020-01-21
chore(refactor) move instagram feed template into reusable component

### LINKED ISSUE #79: Extract instagram feed component
Currently this component is duplicated in 2 pages (category, home):

image

We should extract the one from category page into `InstagramPhotos` component in `src/components`

EugeneHlushko:
- get the markup out into separate file
- use the extracted markup in home and product pages
- Closes #79 
- use vsfui 0.5.1 instead of local version

### DIFF
--- a/packages/commercetools/theme/components/InstagramFeed.vue
+++ b/packages/commercetools/theme/components/InstagramFeed.vue
@@ -1,0 +1,65 @@
+<template>
+  <SfSection title-heading="Share Your Look" subtitle-heading="#YOURLOOK" class="section">
+    <div class="grid grid-images">
+      <div class="grid__row">
+        <div class="grid__col">
+          <SfImage src="/homepage/imageA.jpg">katherina_trn</SfImage>
+        </div>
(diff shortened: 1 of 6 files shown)


packages/commercetools/theme/components/InstagramFeed.vue
```diff
@@ -0,0 +1,65 @@
+<template>
```
filrak, line 1: How about putting it in default layout and then just specifying different layout for checkout? In that way we could avoid repetition)

packages/commercetools/theme/pages/Category.vue
```diff
@@ -491,6 +491,7 @@ export default {
   &__list {
     display: flex;
     flex-wrap: wrap;
+    margin-top: 1.875rem - 0.5rem;
```
filrak, line 494: I actually changed that line in a previous PR
```suggestion
    margin-top: 0;
```

filrak (approved):

Issue #120, opened 2020-01-17
I want to select the country


rmakara:
**Note:** Consider & analyse the use of country vs language vs store in commercetools.

**Acceptance Criteria**
* I can see the list of available countries in a drop-down list.
* Storefront UI component for country selector is used - SfVar.
* Country Selector in `next` project looks like in Figma designs in Desktop and Mobile view.
* Update country in session / context / api. 

**Next tasks**
* #121 Translations

PR #133, opened 2020-01-21
implemented logic for useLanguage composable

defudef:
Related to #120


packages/commercetools/composables/src/useLocale/index.ts
```diff
@@ -0,0 +1,42 @@
...
+      { value: 'AT', label: 'Austria' },
+      { value: 'DE', label: 'Germany' },
+      { value: 'NL', label: 'Netherlands' }
+    ]
+  }
+
+  return {
+    setLanguage,
+    getLanguage,
+    setCountry,
```
andrzejewsky: you don't need a getter for this, just return these values `locale`, `country`, `currency`

andrzejewsky: and for all of available countries / languages please create another field

packages/commercetools/theme/components/AppTopBar.vue
```diff
@@ -0,0 +1,154 @@
...
+export default {
+  setup(props, { isServer }) {
+    const { getLanguage, setLanguage, getCountry, setCountry, getCountries } = useLocale()
+    const countries = reactive([])
+
+    let selectedLanguage = '';
+    let selectedCountry = '';
+
+    if (!isServer) {
+      selectedLanguage = (localStorage.getItem('language') || 'en').toLowerCase()
```
andrzejewsky: that will cause a mismatch (probably) - imagine that on the server you have English but on the client-side you have saved German. In that case, a response that comes from the server is in English but every next request is in German (because you just changed lang)

packages/commercetools/theme/components/AppTopBar.vue
```diff
@@ -0,0 +1,154 @@
...
+      languages: [
+        { value: 'en', label: 'English' },
+        { value: 'de', label: 'Deutsch' },
+      ]
+    }
+  },
+  components: {
+    SfSelect
+  },
+  methods: {
```
andrzejewsky, line 80: just just `setup` and nothing else

packages/core/interfaces/index.ts
```diff
@@ -106,6 +106,14 @@ export interface UseContent<CONTENT, SEARCH> {
   error: any
 }
 
+export interface UseLocale<SET_LANGUAGE, GET_LANGUAGE, SET_COUNTRY, GET_COUNTRY, GET_COUNTRIES> {
+  setLanguage: SET_LANGUAGE,
+  getLanguage: GET_LANGUAGE,
+  setCountry: SET_COUNTRY,
+  getCountry: GET_COUNTRY,
+  getCountries: GET_COUNTRIES
```
andrzejewsky: `setLanguage`, `setCountry`, `country`, `availableCountries`, `language`, `availableLanguages`

andrzejewsky: or - even better, use just `country` and `country` and use reactivity - when I assign a new lang (`language.value = 'en') it will be automatically populated

andrzejewsky (changes requested):

packages/commercetools/theme/components/AppTopBar.vue
defudef: Yes, that's what I've mentioned in previous review. It's temporarily. Once we'll have ready SSR, we'll move local-storage for cookies so it will be supported on SSR and CSR side as well

packages/commercetools/api-client/src/index.ts
```diff
@@ -29,18 +29,23 @@ const setup = <TCacheShape>(setupConfig?: SetupConfig<TCacheShape>): ApolloClien
...
   })
   locale = setupConfig.locale
-  currency = setupConfig.currency
+  currency = (setupConfig.currency || '').toUpperCase()
   country = setupConfig.country
 
   return apolloClient
 }
 
+const updateStoreConfig = (storeConfig: StoreConfig) => {
```
filrak, line 39: imho we should save config passed to `setup` method in `setup` variable and make it exportable. Then we can easily update it's certain properties without unnecessary abstractions like this method.

packages/commercetools/api-client/src/types/setup.ts
```diff
@@ -17,10 +17,15 @@ export interface Token {
   token_type: string
 }
 
-export interface SetupConfig<TCacheShape> {
+export interface StoreConfig {
```
filrak, line 20: According to above we won't need this

packages/commercetools/api-client/src/index.ts
```diff
@@ -29,18 +29,23 @@ const setup = <TCacheShape>(setupConfig?: SetupConfig<TCacheShape>): ApolloClien
     ...setupConfig.customOptions
   })
   locale = setupConfig.locale
-  currency = setupConfig.currency
+  currency = (setupConfig.currency || '').toUpperCase()
```
filrak, line 33: should be renamed to `defaultLocale` and `defaultCurrency`

packages/commercetools/api-client/src/index.ts
```diff
@@ -20,7 +20,7 @@ import getStorage from './helpers/createCommerceToolsLink/getStorage'
 let apolloClient: ApolloClient<any> = null
 let locale = 'en'
 let currency = 'USD'
-let country = 'UK'
```
filrak, line 23: should be nulls at start

packages/commercetools/composables/src/useLocale/index.ts
filrak: for getters I agree, it should be just `country`, `availableCountries` etc, for setter IMHO they should stay as functions are doing more than just assigning variable

packages/commercetools/theme/components/AppTopBar.vue
```diff
@@ -0,0 +1,154 @@
...
+      selectedLanguage,
+      selectedCountry,
+      countries: getCountries(),
+      setLanguage,
+      getLanguage,
+      setCountry,
+      getCountry,
+    }
+  },
+  data () {
```
filrak: should be a part of `setup`

packages/commercetools/theme/components/AppTopBar.vue
filrak, line 80: all of the content in this `methods` will be handled by `nuxt-i18n` - could you use this module instead to not duplicate the logic?

packages/commercetools/composables/src/useLocale/index.ts
```diff
@@ -0,0 +1,42 @@
...
+  }
+  const getLanguage: GetLanguage = () => locale
+  const setCountry: SetCountry = (countryCode: string) => updateStoreConfig({
+    locale,
+    currency,
+    country: countryCode
+  })
+  const getCountry: GetCountry = () => country
+  // TODO: Mocked country list. Need to do: async GQL request
+  const getCountries: GetCountries = () => {
```
filrak: should load automatically when composable is used into `availableCountries`

packages/commercetools/theme/components/AppTopBar.vue
```diff
@@ -0,0 +1,154 @@
+<template>
+  <div class="sf-topbar">
+    <div class="sf-topbar__content">
+      <div class="sf-topbar__content__item"></div>
+      <div class="sf-topbar__content__item">Download our application. <a href="#">Find out more.</a></div>
+      <div class="sf-topbar__content__item selector-container">
+        <SfSelect
+          v-model="selectedLanguage"
```
filrak: it's either v-model or prop/event. `v-model` under the hood runs `change`

packages/commercetools/theme/components/AppTopBar.vue
```diff
@@ -0,0 +1,154 @@
...
+          v-for="language in languages"
+          :key="language.value"
+          :value="language.value"
+        >
+          {{language.label}}
+        </SfSelectOption>
+      </SfSelect>
+      <SfSelect
+          v-model="selectedCountry"
+          @change="onCountrySelect"
```
filrak, line 23: same as above

filrak (changes requested):

PR #153, opened 2020-01-31
Registering

andrzejewsky:


packages/commercetools/composables/src/useUser/index.ts
```diff
@@ -0,0 +1,35 @@
+import { ref, Ref } from '@vue/composition-api'
+import { UseUser } from '@vue-storefront/interfaces'
+import { CustomerSignMeInDraft, CustomerSignMeUpDraft, Customer } from '@vue-storefront/commercetools-api/lib/src/types/GraphQL'
+import  { customerSignMeUp } from '@vue-storefront/commercetools-api'
+
+type UserRef = Ref<Customer>
+type RegisterFn = (userData: CustomerSignMeInDraft) => Promise<void>
```
filrak, line 7: perhaps we could have a more generic type for that and transform to draft format inside composable?

packages/commercetools/theme/components/LoginModal.vue
```diff
@@ -25,104 +25,142 @@
...
+    SfCheckbox,
+    SfLoader,
+    ValidationProvider,
+    ValidationObserver
+  },
   setup() {
+    const form = ref({})
+    const isLogin = ref(false)
+    const createAccount = ref(false)
+    const rememberMe = ref(false)
```
filrak, line 143: imho it should be a param of login form and sent via `login`, wdyt?

packages/commercetools/theme/nuxt.config.js
```diff
@@ -98,6 +98,9 @@ export default {
     endpoint: 'https://lovecrafts-dev.cdn.prismic.io/api/v2'
   },
   build: {
+    transpile: [
```
filrak, line 101: Why we need that? Maybe add this to nuxt module (just do `if` checking if thats installed) same way as we have with sfui transpilation now

andrzejewsky: https://logaretm.github.io/vee-validate/guide/rules.html#importing-rules-in-nuxt-js

packages/commercetools/theme/components/LoginModal.vue
andrzejewsky: it is, it displays only on the login page, I used separated field because there are only fields that I needed to the registering - next task is for login I'll handle that

Issue #149, opened 2020-01-31
Change Prismic library


rmakara:
https://github.com/prismicio/prismic-vue

@filrak to describe the task

PR #155, opened 2020-02-04
#149 Prismic as a package

defudef:


packages/commercetools/theme/pages/Prismic.vue
```diff
@@ -1,14 +1,32 @@
...
+import { usePrismic } from '@vue-storefront/prismic'
+import PrismicDocument from '@vue-storefront/prismic/components/PrismicDocument'
 
 export default {
   components: {
-    CustomPage,
+    PrismicDocument
+  },
+  setup() {
+    const { document, search, loading, error } = usePrismic()
```
andrzejewsky: `document` is a JS keyword, use something different instead, eg. `doc`

packages/prismic/components/PrismicBlock.vue
```diff
@@ -0,0 +1,82 @@
+<template>
+  <div v-html="renderBlock()" />
+</template>
+
+<script>
+import PrismicDom from 'prismic-dom'
+import { usePrismic } from '@vue-storefront/prismic'
+
+const parseType = (block) => {
```
andrzejewsky: transform these functions into internal helpers

andrzejewsky: and add TS types for them & write test cases

packages/prismic/components/PrismicBlock.vue
```diff
@@ -0,0 +1,82 @@
...
+  props: {
+    block: {
+      type: Object | String | Number,
+      required: true
+    }
+  },
+  setup({ block }) {
+    const { render } = usePrismic()
+
+    const { type, document } = transformBlock(block)
```
andrzejewsky: this should be part of the composable... 
```
const { document } = usePrismic()
```

and the `render` or `renderBlock` - sounds like public helper (`getPrismicBlock`)

andrzejewsky: import { getPrismicBlock } from '@vue-storefront/prismic/helpers'
import { usePrismic } from '@vue-storefront/prismic/composables'

```
setup() {
   const { document } = usePrismic()

  return { document, getPrismicBlock}
}
```
and template:

```
<tepmlate>
<div>{{ getPrismicBlock(document) }}</div>
</template>
```

andrzejewsky: + ofc unit test for the helper

packages/prismic/components/PrismicDocument.vue
```diff
@@ -0,0 +1,73 @@
...
+  <div>
+    <prismic-block :key="block.key" v-for="block in blocks" :block="block.element" />
+  </div>
+</template>
+
+<script>
+import { reactive } from '@vue/composition-api'
+import PrismicBlock from '@vue-storefront/prismic/components/PrismicBlock'
+
+const parseType = (document) => {
```
andrzejewsky: exactly the same in here: internal helper, unit tests, etc.

packages/prismic/src/composables/usePrismic/index.ts
```diff
@@ -0,0 +1,87 @@
...
+      if (!prismic.Predicates[key]) {
+        return undefined
+      }
+
+      const { fragment, after, before, day, documentId, 
+        hour, latitude, longitude, maxResults, month, 
+        radius, value, values, year 
+      } = current
+
+      switch (true) {
```
andrzejewsky: ????

packages/prismic/src/interfaces.ts
```diff
@@ -0,0 +1,105 @@
+type DateValue = string | number | Date
```
andrzejewsky: directory types/index.ts as in others

andrzejewsky (changes requested):

packages/prismic/src/composables/usePrismic/index.ts
defudef: Instead of having tons of ifs I've written switch true. I've seen this approach several times

packages/prismic/components/PrismicBlock.vue
defudef: I can agree `document` isn't the correct name for the block content. `element` sounds better so I'll change it.

Also, every other package uses helper only inside composables so we should keep a convention.

The important thing is, I don't want to share the rendering method "globally". This is only a "ready to go" solution. You can but you're not forced to render every block in this way.

EugeneHlushko (commented):
Hi @defudef 
It would be nice to get it documented or extend PR description with key points so its easier to review, from code perspective it looks good despite slices and `doc.body` being a TODO :)

defudef:
> Hi @defudef
> It would be nice to get it documented or extend PR description with key points so its easier to review, from code perspective it looks good despite slices and `doc.body` being a TODO :)

Working on that now as separate PR :)

andrzejewsky (approved):

PR #161, opened 2020-02-06
core(theme) Theme inheritance PoC 

filrak:
Once there is a file in `/app/{dir}` it overrides `nuxt-module/theme/{dir}`


packages/core/theme-module/index.js
```diff
@@ -0,0 +1,68 @@
...
+
+const log = {
+  info: (message) => consola.info(chalk.bold('VSF'), message),
+  success: (message) => consola.success(chalk.bold('VSF'), message),
+  warning: (message) => consola.warning(chalk.bold('VSF'), message),
+  error: (message) => consola.error(chalk.bold('VSF'), message)
+}
+
+const getAllFiles = function(dirPath) {
+  files = fs.readdirSync(dirPath)
```
defudef: missing const

packages/core/theme-module/index.js
```diff
@@ -0,0 +1,68 @@
...
+ 
+  arrayOfFiles = []
```
defudef: missing let

packages/core/theme-module/index.js
```diff
@@ -0,0 +1,68 @@
...
+ 
+  files.forEach(function(file) {
```
defudef: Just use map:

```JS
files.map(file => {
   if (fs.statSync(`${dirPath}/${file}`).isDirectory()) {
     return ...getAllFiles(dirPath + '/' + file, arrayOfFiles)
   } else {
      return`${dirPath}/${file}`.split(__dirname + '/').pop()
})
```

packages/core/theme-module/index.js
andrzejewsky: he wants here a global one : )

packages/core/theme-module/index.js
andrzejewsky: I would say reduce...  but... it's just a proposition:
```
return files.reduce((prev, curr) => {
  if (fs.statSync(dirPath + "/" + curr).isDirectory()) {
    return getAllFiles(dirPath + "/" + curr, prev)
  }

  const newFiles = (dirPath + "/" + curr).split(__dirname + '/').pop()

  return [...prev, ...newFiles]
}, [])
```

andrzejewsky (approved):
remove error log

PR #167, opened 2020-02-13
eslint and precommit hook setup

Qrzy:
Initial setup for eslint and precommit hook linting staged changes.

### DIFF
--- a/package.json
+++ b/package.json
@@ -26,15 +26,22 @@
     "dev:ct:helpers": "cd packages/commercetools/helpers && yarn dev",
     "dev:ay:api-client": "cd packages/about-you/api-client && yarn dev",
     "dev:ct": "cd packages/commercetools/theme && yarn dev",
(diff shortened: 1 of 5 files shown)


package.json
```diff
@@ -26,15 +26,23 @@
     "dev:ct:helpers": "cd packages/commercetools/helpers && yarn dev",
     "dev:ay:api-client": "cd packages/about-you/api-client && yarn dev",
     "dev:ct": "cd packages/commercetools/theme && yarn dev",
-    "heroku-postbuild": "yarn build:ct"
+    "heroku-postbuild": "yarn build:ct",
```
andrzejewsky: can you delete that? and the Procfile - we are not using it anymore

andrzejewsky (approved):

package.json
Qrzy: sure, on my way

PR #173, opened 2020-02-18
add related products

Qrzy:


packages/core/theme-module/theme/components/RelatedProducts.vue
```diff
@@ -0,0 +1,75 @@
...
+
+  props: {
+    title: String,
+    product: Object,
+  },
+
+  setup({ product }) {
+    const { products, search, loading, error } = useProduct();
+    const categories = computed(() => product ? getProductCategories(product) : [])
+    const displayProducts = computed(() => getProductVariants(products.value, { masters: true }).filter(prod => getProductId(prod) !== getProductId(product)))
```
andrzejewsky, line 54: It might be useful to hide this filtering inside of the `getProductVariant` helper, eg. we can introduce another parameter eg. `exclude: { id }` and `exclude` means just array.filter. @filrak what do you think?

packages/core/theme-module/theme/pages/Product.vue
```diff
@@ -168,36 +168,11 @@
...
-            class="product-card"
-          />
-        </SfCarouselItem>
-      </SfCarousel>
-    </SfSection>
+    <RelatedProducts
+      v-if="product"
+      :product="product"
+      title="Match it with"
+    />
```
andrzejewsky, line 175: short explanations for the rest of the team: I suggested to display just one section of related products instead of two because we had no idea what to display, what searching criteria we can use - the first section displays just products from the same category as a product you visited

andrzejewsky (changes requested):

packages/commercetools/helpers/src/index.ts
```diff
@@ -86,6 +91,9 @@ export const getProductAttributes = (products: ProductVariant[], filterByAttribu
 
 export const getProductDescription = (product: ProductVariant): any => (product as any)._description
 
+export const getProductCategories = (product: ProductVariant): string[] => (product as any)._categoriesRef
+
+export const getProductId = (product: ProductVariant): number => (product as any)._id
```
filrak, line 96: I don't think every platform has "id"

filrak (changes requested):

packages/commercetools/helpers/src/index.ts
andrzejewsky: @filrak but most likely yes (magento, ct, about-you - they all have id), but it could be another identifier, while this helper can have the still same name.

andrzejewsky (approved):

packages/core/theme-module/theme/components/RelatedProducts.vue
filrak, line 54: why not `relatedProducts` to be more descriptive

filrak (changes requested):

packages/core/theme-module/theme/components/RelatedProducts.vue
```diff
@@ -0,0 +1,75 @@
...
+
+    watch(categories, () => {
```
filrak, line 56: Do we really need that? Can't we just show other products from this category and thats it?

PR #184, opened 2020-02-20
#183 Covered edge cases, support primary slice

defudef:


packages/prismic/src/composables/usePrismic/index.ts
```diff
@@ -1,36 +1,65 @@
...
+    loading.value = true;
+
+    await prismic
+      .getApi(endpoint, apiOptions)
+      .then((api) => api.query(
+        transformQuery(query),
+        options as QueryOptions
+      ))
+      // eslint-disable-next-line
+      .then(({ results, total_results_size, total_pages, results_size, results_per_page, prev_page, page, next_page }) => {
```
andrzejewsky: I totally don't get it why you reading all of the properties and also you need an additional field `meta` while you already created helpers for reading properties such as total_pages, results_size etc. can't you assign entire document to the `doc`?

packages/prismic/src/composables/usePrismic/index.ts
```diff
@@ -1,36 +1,65 @@
...
+          total_pages, // eslint-disable-line
+          total_results_size // eslint-disable-line
+        };
+        doc.value = results;
+      });
+
+    loading.value = false;
+  };
+
+  const searchFirst: Search = async (query: PrismicQuery | PrismicQuery[], options: OptionsType = {}) => {
```
andrzejewsky: it should be just`search` function and nothing else if the result from the API is different just use different arguments to manipulate the returning value

packages/prismic/src/composables/usePrismic/transformQuery.ts
```diff
@@ -43,11 +44,11 @@ const transformQuery = (query: PrismicQuery): string | string[] => {
       default:
         return predict(key, [fragment]);
     }
-  });
+  }));
+
+  const flatArray = [].concat(...queries.filter((queryElement) => queryElement !== undefined));
```
andrzejewsky, line 49: `const flatArray = queries.filter((queryElement) => queryElement);`

andrzejewsky (changes requested):

packages/prismic/src/composables/usePrismic/index.ts
```diff
@@ -1,36 +1,65 @@
...
 
-  const search: Search = async (query: PrismicQuery, options: OptionsType = {}) => {
+  const search: Search = async (query: PrismicQuery | PrismicQuery[], options: OptionsType = {}) => {
+    loading.value = true;
+
+    await prismic
+      .getApi(endpoint, apiOptions)
+      .then((api) => api.query(
+        transformQuery(query),
+        options as QueryOptions
```
andrzejewsky: if this isn't the OptionsType so maybe you should use QueryOptions instead of OptionType in the search definition?

andrzejewsky (changes requested):

packages/prismic/src/composables/usePrismic/index.ts
defudef: I can't as queryFirst returns document only, while query returns ApiSeachResponse type

packages/prismic/src/composables/usePrismic/transformQuery.ts
defudef: I need to concatenate  it to flatten an array

packages/prismic/src/composables/usePrismic/index.ts
defudef: Refactored a little. As we talked, Prismic's type is not enough, that's why I've added another one

andrzejewsky (approved):

PR #213, opened 2020-02-24
add orders history to my account

### LINKED ISSUE #182: Implement / update useUser composable for loading the orders history

### LINKED ISSUE #190: I want to see the history of my orders (headers) So I can track the progress of its realization
**Context**
After I place many orders in the shop I want to review its history. The part of that task is to display just a table of the historical orders, that I placed. At this stage, it should not be possible to open the details of the orders.

**Acceptance Criteria**
* In My Account I can see the possibility to open My Orders.
* I can see the grid with all the orders, that I place.
* I can see order date, order number, status, total value net, total value vat, total value gross.

**Not in the scope**
* Details view should be implemented in the next task https://github.com/DivanteLtd/next/issues/197. NOT IN THIS ONE.

Qrzy:


andrzejewsky (commented):
@filrak do we need a another field for orders or can we include it in the user field? What do you think? I think having different field just for orders is even better (no enhance functions in the future etc.)

filrak:
https://github.com/DivanteLtd/next/pull/229

packages/commercetools/api-client/src/index.ts
```diff
@@ -19,6 +19,7 @@ import customerSignMeUp from './api/customerSignMeUp';
 import customerSignMeIn from './api/customerSignMeIn';
 import customerSignOut from './api/customerSignOut';
 import getStorage from './helpers/createCommerceToolsLink/getStorage';
+import getMyOrders from './api/getMyOrders';
```
patzick, line 22: this is clearer for index :)
```suggestion
export * from './api/getMyOrders';
```

packages/core/factories/src/useUserOrdersFactory.ts
```diff
@@ -0,0 +1,33 @@
+import { ref, Ref, computed } from '@vue/composition-api';
+import { UseUserOrders } from '@vue-storefront/interfaces';
+
+export type OrdersSearchResult<ORDER> = {
+  data: ORDER[];
+  total: number;
+};
```
patzick, line 7: same as with #265 - common interface for this
```suggestion
export type SearchResults<T> = {
  data: T[];
  total: number;
};
```

packages/core/factories/src/useUserOrdersFactory.ts
```diff
@@ -0,0 +1,33 @@
...
+
+export function useUserOrdersFactory<ORDER, ORDER_SEARCH_PARAMS>(factoryParams: UseUserOrdersFactoryParams<ORDER, ORDER_SEARCH_PARAMS>) {
+  return function useUserOrders(): UseUserOrders<ORDER> {
+    const orders: Ref<OrdersSearchResult<ORDER>> = ref({ data: [], total: 0 });
+    const loading: Ref<boolean> = ref(false);
+
+    const searchOrders = async (params?: ORDER_SEARCH_PARAMS): Promise<void> => {
+      loading.value = true;
+      orders.value = await factoryParams.searchOrders(params);
```
patzick: if this throws an exception, then loading will be on indefinitely

patzick (changes requested):
Looks good, just some minor things here

andrzejewsky (approved):

packages/commercetools/api-client/src/index.ts
Qrzy: I agree, just in this particular case it ends with some unexpected error thus I'd like to handle this change for entire file separately. :)

packages/core/factories/src/useUserOrdersFactory.ts
Qrzy: Good point, I also moved it to interfaces.

PR #215, opened 2020-02-25
#185 Extended usePrismic functionality

defudef:
What left:
- [x] filtering slices optional function
- [x] override transformBlock
- [x] coverage all stuff with unit tests

### DIFF
--- a/packages/prismic/src/helpers/index.ts
+++ b/packages/prismic/src/helpers/index.ts
@@ -1,5 +1,5 @@
-import { transformBlock } from './_utils';
-import { PrismicSlice, PrismicMeta } from '../types';
+import { transform as transformFunction } from '../index';
+import { PrismicSlice, PrismicMeta, TransformBlock } from '../types';
 import { Document } from 'prismic-javascript/d.ts/documents';
 
 export const getPages = (doc: Document | Document[], pageUid?: string): Document | null | Document[] => {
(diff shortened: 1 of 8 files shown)


packages/prismic/src/helpers/index.ts
```diff
@@ -42,44 +42,66 @@ export const getPageSlugs = (page: Document): string[] => page.slugs;
...
+    };
+  }
 
-    return [foundSlice.primary || {}, ...foundSlice.items].map((item) => renderSliceElements(item));
+  const filteredSlices = Array.isArray(sliceType)
+    ? slices.filter((slice) => sliceType.includes(slice.slice_type))
+    : slices;
+
+  if (typeof sliceType === 'function') {
+    return filteredSlices
```
EugeneHlushko: maybe just make it a `let` and overwrite here to have single return e.g.
```suggestion
    filteredSlices =  filteredSlices.filter((slice) => sliceType(slice));
```

packages/prismic/src/helpers/index.ts
```diff
@@ -42,44 +42,66 @@ export const getPageSlugs = (page: Document): string[] => page.slugs;
...
+  return filteredBlockKeys.map((key) => transform(data[key]));
+};
 
+export const getSlices = ({ data }: Document, sliceType?: ElementNameType | FilterSlice): any => {
   const slices = data.body as PrismicSlice[];
 
-  if (sliceType) {
-    const foundSlice = slices.find((slice) => slice.slice_type === sliceType);
+  if (typeof sliceType === 'string') {
+    const foundSlices = slices.find((slice) => slice.slice_type === sliceType);
```
EugeneHlushko: This only find single slice of given type, it has to be an array, maybe use `.filter` here as well

packages/prismic/src/helpers/index.ts
```diff
@@ -42,44 +42,66 @@ export const getPageSlugs = (page: Document): string[] => page.slugs;
...
+        primary,
+        items
+      }));
   }
 
-  return slices
-    .map((slice) => [slice.primary || {}, ...slice.items]
-      .map((item) => renderSliceElements(item))
-    );
+  return filteredSlices
```
EugeneHlushko: dont need to map that here just return original slices as discussed in slack

andrzejewsky (approved):

PR #227, opened 2020-03-02
Added readme for prismic lib

defudef:


patzick (approved):
👌

packages/prismic/README.md
```diff
@@ -0,0 +1,83 @@
...
+
+With NPM:
+```bash
+npm install --save @vue-storefront/prismic
+```
+
+## Setup library
+
+```typescript
+import { setup } from '@vue-storefront/prismic'
```
EugeneHlushko, line 20: missing semi

packages/prismic/README.md
```diff
@@ -0,0 +1,83 @@
...
+  endpoint: 'https://your-account.prismic.io/api/v2',
+  apiOptions: {
+    accessToken: 'Prismic_Access_Token'
+  }
+});
+```
+
+## Nuxt support
+
+If you need to integrate Prismic with Nuxt, just simply provide setup file (that one above for example) into `nuxt.config.js` plugins section.
```
EugeneHlushko, line 32: words like Simply, Just can be demotivating for people struggling with issues on these steps, should remove both and it still will read ok. When suggessting configuration, always provide an example e.g.

---------------------------------------------------------------------------------------

If you need to integrate Prismic with Nuxt, create a setup file (shown in example above) and register it in your `nuxt.config.js` plugins section.

__nuxt.config.js__

```javascript
export default {
  // ...
  plugins: ['./path/to/your-prismic-setup-file.js']
}
```


```

**Review by EugeneHlushko — COMMENTED** [review]
This is a good starting point, although missing some detail and polishing

File: packages/prismic/README.md [hunk 3]
```diff
@@ -0,0 +1,83 @@
...
+      }
+    });
+
+    return {
+      pages,
+      loading,
+      error,
+      getBlocks,
+      getSlices,
+      getPageId
```
EugeneHlushko, line 78: Would be nice to document all of these separately

PR #228, opened 2020-03-02
Feature/introduce boilerplate theme

I2olanD:
Hey,

i added the boilerplate theme. the build is still failing since there are some dependency issues inside the nuxt-theme which i could not figure out. Somehow the dependency path is not updated correctly in the nuxt-theme while building. In the .nuxt folder i still find components that require commerce-tool dependencies. 

I did a lerna version upgrade and added some required dependencies to a couple packages. I also removed hoisting dependencies as i think, that all of the packages should work independently?

Updated also the readme, maybe it's a bit more clear how to set everything up and get started faster :).


lerna.json
```diff
@@ -1,4 +1,11 @@
 {
-  "useWorkspaces": "true",
```
filrak: Changing the way how lerna works in the project was not part of the issue

packages/boilerplate/theme/.babelrc
```diff
@@ -0,0 +1,16 @@
+{
```
filrak: Why do we need that config in a theme?

packages/boilerplate/theme/.editorconfig
```diff
@@ -0,0 +1,13 @@
+# editorconfig.org
```
filrak, line 1: That wasn't in a scope of this issue but we can keep it;)

packages/boilerplate/theme/jest.config.js
```diff
@@ -0,0 +1,19 @@
+const baseConfig = require('./../jest.base.config');
```
filrak: jest config is not needed in nuxt

packages/boilerplate/theme/plugins/api-client.js
```diff
@@ -0,0 +1,4 @@
+import { setup } from '@vue-storefront/about-you-api';
```
filrak:
```suggestion
import { setup } from '@vue-storefront/boilerplate-api';
```

packages/boilerplate/theme/test/pages/Product.spec.js
```diff
@@ -0,0 +1,4 @@
+/* eslint-disable camelcase, @typescript-eslint/camelcase */
```
filrak: was not a scope of this issue, we will do e2e for that part

packages/boilerplate/theme/tsconfig.json
```diff
@@ -0,0 +1,34 @@
+{
```
filrak, line 1: ideally we should use config from tsnuxt (like in CT theme)

packages/commercetools/theme/package.json
```diff
@@ -20,7 +20,10 @@
     "@vue-storefront/nuxt": "^0.0.3",
     "@vue-storefront/nuxt-theme": "^0.0.1",
     "@vue-storefront/prismic": "^0.0.1",
+    "@vue-storefront/utils": "^0.0.3",
```
filrak: i dont think its needed in theme, only in composables pckg

packages/commercetools/theme/package.json
```diff
@@ -20,7 +20,10 @@
...
+    "@vue/composition-api": "^0.4.0",
     "core-js": "^2.6.5",
+    "loglevel": "^1.6.7",
```
filrak: i dont think its needed here, if so its added by nuxt pckg

packages/boilerplate/helpers/package.json
```diff
@@ -18,5 +18,6 @@
   },
```
filrak: whole helpers package will be deleted anyway

packages/boilerplate/theme/pages/Product.vue
```diff
@@ -0,0 +1,11 @@
+<template>
```
filrak: we dont keep pages in integration-specific themes

filrak (changes requested):

filrak (approved):

PR #230, opened 2020-03-03
refactor: convert helpers function to getters

krskibin:
# 2020-03-11 update
Due to performance and stability problems, we agreed that our priority is to make our codebase as simple as possible and also keep the highest performance in the mobile application. For this reason, we decided that the returned "Helper functions" from `from` would not be based on the passed to itself argument. Instead of this, each composable will have access to a Helper functions prepared especially for them.

Most common usage:
```js
import useProduct from '@vue-storefront/commercetools';
setup() {
  const { helpers, product } = useProduct();
  return {
    name: helpers.getName(product);
  }
}
```
with two diffrent composables:
```js
import useProduct from '@vue-storefront/commercetools';
import useCart from '@vue-storefront/commercetools';

setup() {
  const { productHelpers: helpers, product } = useProduct();
  const { cartHelpers: helpers, product } = useCart();
  
  return {
    name: helpers.getName(product.value);
    cartProducts: cartHelpers.getProducts(cart.value)
  }
}
```


packages/boilerplate/composables/src/getters/_utils.ts
```diff
@@ -0,0 +1,5 @@
+import {isRef, ref, Ref} from '@vue/composition-api';
+
+export function unwrap<T>(element: Ref<T> | T): Ref<T> {
```
Qrzy: This is rather `wrap` than `unwrap` since it's wrapping element with `Ref` if it's not wrapped already.

packages/boilerplate/composables/src/getters/cartHelpers.ts
```diff
@@ -0,0 +1,23 @@
+/* eslint-disable no-unused-vars, @typescript-eslint/no-unused-vars */
```
Qrzy: Should those files be called `...Helpers`? This doesn't conform to parent directory name, `getters`. 
Would you consider just removing the `Helpers` postfix from filename?

packages/boilerplate/composables/src/getters/index.ts
```diff
@@ -0,0 +1,16 @@
+// TODO: Create getters import.
+import * as productHelpers from './productHelpers';
```
Qrzy: And then also a change here would be needed, `productHelpers` -> `productGetters`. 
What do you think?

packages/boilerplate/composables/src/getters/cartHelpers.ts
filrak: +1 for getters

packages/commercetools/composables/src/useUser/index.ts
```diff
@@ -49,6 +50,18 @@ export default function useUser(): UseUser<Customer, any> {
     loading.value = false;
   });
 
+  const userGetters = {
```
filrak, line 55: talk to Konrad about that

filrak: Also the typesc here should come from the factory, same as it is with other composable properties

packages/core/theme-module/theme/components/CartSidebar.vue
```diff
@@ -13,20 +13,20 @@
             <transition-group name="fade" tag="div">
               <SfCollectedProduct
                 v-for="product in products"
-                :key="getCartProductName(product)"
-                :image="getCartProductImage(product)"
-                :title="getCartProductName(product)"
-                :regular-price="getCartProductPrice(product)"
+                :key="cartGetters.getProductName(product).value"
```
filrak: i think it will look cleaner if you will return it like this:
```
return {
  productName: cartGetters.getProductName(product)
}
```
- you will see imemdiately what is available in a template
- no need to use ugly `value`

packages/core/theme-module/theme/components/CartSidebar.vue
```diff
@@ -103,10 +92,10 @@ export default {
     SfCollectedProduct
   },
   setup() {
-    const { cart, removeFromCart, updateQuantity } = useCart();
-    const products = computed(() => getCartProducts(cart.value));
-    const totals = computed(() => getCartTotals(cart.value));
-    const totalItems = computed(() => getCartTotalItems(cart.value));
+    const { cartGetters, cart, removeFromCart, updateQuantity } = useCart();
+    const products = cartGetters.getProducts(cart);
```
filrak: same, can be in return statemant

packages/core/theme-module/theme/components/RelatedProducts.vue
```diff
@@ -61,10 +52,10 @@ export default {
       relatedProducts,
       search,
       loading,
-      getProductSlug,
-      getProductName,
-      getProductGallery,
-      getProductPrice
+      getProductSlug: product => productGetters.getSlug(product),
```
filrak, line 55: why its a function? also its not a function so it should be `productSlug`

packages/core/theme-module/theme/components/TopBar.vue
```diff
@@ -37,14 +36,13 @@ export default {
...
-    const { isAuthenticated, logout, user } = useUser();
+    const { isAuthenticated, logout, user, userGetters } = useUser();
 
     return {
       toggleLoginModal,
       isAuthenticated,
       logout,
-      user,
-      getUserFullName
+      userFullName: userGetters.getFullName(user)
```
filrak, line 45: this is the right usage ;)

packages/core/theme-module/theme/components/checkout/CartPreview.vue
```diff
@@ -135,34 +120,28 @@ export default {
...
-      getShippingMethodName,
-      getShippingMethodDescription,
-      getShippingMethodPrice,
-      getCartProductName,
-      getCartProductImage,
-      getCartProductPrice,
-      getCartProductQty,
-      getCartProductAttributes,
-      getCartProductSku,
+      getCartProductName: product => cartGetters.getProductName(product).value,
```
filrak: all here should be just
`cartProductName: cartGetters.getProductName(product)`

packages/core/theme-module/theme/pages/Category.vue
```diff
@@ -268,29 +260,27 @@ export default {
...
-      products,
-      productsLoading,
       categoryTree,
-      getProductName,
-      getProductGallery,
-      getProductPrice,
-      getProductSlug,
+      productsLoading,
+      products: productGetters.getVariants(categoryProducts, { master: true}),
+      getProductName: product => productGetters.getName(product).value,
```
filrak: same

packages/core/theme-module/theme/pages/Checkout/Shipping.vue
```diff
@@ -163,18 +156,25 @@ export default {
...
+      getShippingMethodId,
+      getShippingMethodName
+    } = checkoutGetters;
 
     return {
       shippingDetails,
       chosenShippingMethod,
       shippingMethods,
-      getShippingMethodName,
       getShippingMethodDescription,
```
filrak, line 176: no `value` in ui, and that should be in the form suggested previously

packages/core/theme-module/theme/pages/Product.vue
```diff
@@ -222,51 +222,39 @@ import {
...
-
     return {
       updateFilter,
       configuration,
       product,
       options,
-      getProductName,
-      getProductPrice,
-      getProductGallery,
+      getProductName: product => productGetters.getName(product).value,
```
filrak: same

filrak (changes requested):
Overally outstanding job! Congrats!

Just keep consistency in how you're exposing getters to the UI

packages/boilerplate/composables/src/getters/_utils.ts
krskibin: Yes, thats right ;P

packages/boilerplate/composables/src/getters/cartHelpers.ts
krskibin: Yes, I decided to add helpers at the end, because of the naming conflict with the getters from the composables, but you right, maybe I would change filenames to just `product, cart, ...etc` and in composables to `import {productStatics (productFn)} from './gettes'`. What do you think?

krskibin: With @filrak we decided to rename it to `productPlainGetters`

packages/boilerplate/composables/src/getters/index.ts
krskibin: I will change it to `productPlainGetters`, as I wrote in comment above.

packages/commercetools/composables/src/useUser/index.ts
krskibin: We will refactor the useUser after Konards changes will be merged, for now, I think we can keep it as it is.

packages/boilerplate/composables/src/getters/cartHelpers.ts
patzick: I believe filename should be same as getters:
`productGetters -> geetters/productGetters.ts`
easy to find when you're interested with that particular getter. Just `product` wouldn't be enough and `productPlainGetters` wouldn't be any clue

packages/boilerplate/composables/src/getters/cartHelpers.ts
```diff
@@ -0,0 +1,23 @@
...
+import { Cart } from '@vue-storefront/boilerplate-api/src/types';
+
+// TODO: Change UICartProduct to something agnostic.
+export const getProducts = (cart: Cart, includeAttributes: string[] = []): any[] => {
```
patzick: includeAttributes should be in object. Overall
`getSomething(source, options)`
includeAttributes is in this case one of the properties here.
Also should be optional

packages/commercetools/composables/src/getters/cartHelpers.ts
```diff
@@ -0,0 +1,51 @@
+import { productHelpers } from './index';
```
patzick:
```suggestion
import { getAttributes } from './productGetters';
```

packages/boilerplate/composables/src/getters/index.ts
```diff
@@ -0,0 +1,14 @@
...
+import * as cartHelpers from './cartHelpers';
+import * as categoryHelpers from './categoryHelpers';
+import * as checkoutHelpers from './checkoutHelpers';
+import * as userHelpers from './userHelpers';
+
+export default {
```
patzick: With default export, there is no tree-shaking here. Please change the lines above as an example:
```ts
export * as userHelpers from './userHelpers';
```

packages/commercetools/composables/src/useUser/index.ts
patzick: this shouldn't be here, just something at return like
```
return {
  // props
  productGetters: makeComputedGetters(productGetters)
}
```
otherwise it's an implementation nightmare :)

packages/core/factories/src/useCartFactory.ts
```diff
@@ -27,6 +28,19 @@ export type UseCartFactoryParams<CART, CART_ITEM, PRODUCT, COUPON> = {
...
+    getProductName: (product: PRODUCT) => string;
+    getProductImage: (product: PRODUCT) => string;
+    getProductPrice: (product: PRODUCT) => string;
+    getProductQty: (product: PRODUCT) => string;
+    getProductAttributes:
+      (product: PRODUCT, filterByAttributeName?: Array<string>) => Record<string, string | AgnosticProductAttribute>;
+    getProductSku: (product: PRODUCT) => string;
+    getTotals: (cart: CART) => AgnosticTotals;
+    getShippingPrice: (cart: CART) => number;
+    getTotalItems: (cart: CART) => number;
```
patzick: we need to allow here to add more helpers

patzick (changes requested):
Not reviewed all yet, leeft few comment and waiting for "to review" state :)
Just please remember: 
- if there is any additional param in getter needed, then should be as an attribute of second `options` param, which is an optional object

packages/core/factories/src/useCartFactory.ts
krskibin: Now, this type is inferred core/interfaces

packages/core/interfaces/src/index.ts
```diff
@@ -158,6 +165,60 @@ export interface UseLocale
...
+    ComputedProperty<PRODUCT[]> | Readonly<PRODUCT[]>;
+  getAttributes: (product: PRODUCT[] | Readonly<PRODUCT[]>, filters?: Array<string>) =>
+    ComputedProperty<Array<AgnosticProductAttribute>>;
+  getDescription: (product: PRODUCT | Readonly<PRODUCT>) => ComputedProperty<any>;
+  getCategories: (product: PRODUCT | Readonly<PRODUCT>) => ComputedProperty<string[]>;
+  getId: (product: PRODUCT | Readonly<PRODUCT>) => ComputedProperty<number>;
+}
+
+export interface CartGetters<CART, PRODUCT> {
+  getProducts: (cart: CART) => ComputedProperty<PRODUCT[]>;
```
mkucmus, line 185: Shopware and many other commerce platforms have a CartItem entity instead of Product directly.
Though the product is the most obvious type of cart item, there is something like a coupon, etc. 
We just need to keep in mind that fact. Maybe a little bit deeper level of abstraction is needed here? We are going to face this problem shortly ;)

packages/core/interfaces/src/index.ts
```diff
@@ -158,6 +165,60 @@ export interface UseLocale
   loading: ComputedProperty<boolean>;
 }
 
+export interface ProductGetters<PRODUCT, PRODUCT_FILTER> {
```
mkucmus, line 170: maybe including the ProductType (simple, parent, etc) is a good way to recognize which product we are dealing with,
so `getType` method is maybe a solution here.

packages/core/interfaces/src/index.ts
```diff
@@ -158,6 +165,60 @@ export interface UseLocale
...
+  [getterName: string]: (element: CATEGORY, ...args: any) => any;
+}
+
+export interface UserGetters<USER> {
+  getFirstName: (customer: USER | Ref<USER>) => ComputedProperty<string>;
+  getLastName: (customer: USER | Ref<USER>) => ComputedProperty<string>;
+  getFullName: (customer: USER | Ref<USER>) => ComputedProperty<string>;
+}
+
+export interface CheckoutGetters<SHIPPING_METHODS> {
```
mkucmus, line 210: what about BillingMethod stuff?

packages/core/interfaces/src/index.ts
```diff
@@ -158,6 +165,60 @@ export interface UseLocale
...
+  getShippingMethodId: (shippingMethod: SHIPPING_METHODS) => ComputedProperty<string>;
+  getShippingMethodName: (shippingMethod: SHIPPING_METHODS) => ComputedProperty<string>;
+  getShippingMethodDescription: (shippingMethod: SHIPPING_METHODS) => ComputedProperty<string>;
+  getShippingMethodPrice: (shippingMethod: SHIPPING_METHODS) => ComputedProperty<number>;
+}
+
+export interface UserOrderGetters<ORDER> {
+  getDate: (order: ORDER) => ComputedProperty<string>;
```
mkucmus, line 218: `getItems` to get the ordered lines of products/items (including things like totals have - for instane, shipping price line)

packages/core/interfaces/src/index.ts
mkucmus: the second thing is that many (magento, shopware for sure) ecommerce platforms have related products set on particular Product entity as one to many relation. Maybe we need to have a method like `getRelatedProducts` method as well.

packages/core/interfaces/src/index.ts
```diff
@@ -158,6 +165,60 @@ export interface UseLocale
   loading: ComputedProperty<boolean>;
 }
 
+export interface ProductGetters<PRODUCT, PRODUCT_FILTER> {
+  getName: (product: PRODUCT | Readonly<PRODUCT>) => ComputedProperty<string>;
+  getSlug: (product: PRODUCT | Readonly<PRODUCT>) => ComputedProperty<string>;
+  getPrice: (product: PRODUCT | Readonly<PRODUCT>) => ComputedProperty<number | null>;
```
mkucmus, line 173: Unless we have to deal with regular price only, we need to return a object of prices (regular: 100$, finalPrice: 80%), or add some other getter for special price - after discount, for example.

packages/core/interfaces/src/index.ts
```diff
@@ -15,6 +15,7 @@ export interface UseProduct<PRODUCT> {
   }) => Promise<void>;
   loading: ComputedProperty<boolean>;
   [x: string]: any;
```
mkucmus, line 17: are we going to not using the error ref in case of problems? (regarding all composables which can do the DAO calls)

filrak (approved):

packages/core/theme-module/theme/components/checkout/CartPreview.vue
filrak: still unresolved

packages/core/interfaces/src/index.ts
krskibin: Sure, there is also a CART_ITEM type, which I can use here.

packages/core/utils/src/makeComputedGetters/index.ts
```diff
@@ -0,0 +1,14 @@
+import {computed} from '@vue/composition-api';
+import wrap from '../wrap';
+
+interface Getters<T> {
+  [key: string]: (param: T, ...args: any) => unknown;
```
patzick: the return type is not unknown, it's Readonly<unknown> :)

packages/core/utils/__tests__/wrap.spec.ts
```diff
@@ -0,0 +1,15 @@
+import wrap from '../src/wrap';
+import { ref } from '@vue/composition-api';
+
+describe('[CORE utils] unwrap', () => {
```
patzick:
```suggestion
describe('[CORE utils] wrap', () => {
```

packages/core/utils/__tests__/makeComputedGetters.spec.ts
```diff
@@ -0,0 +1,30 @@
...
+  it('should map all getter functions to computed getters', () => {
+    const mockGetters = {
+      getProducts: param => `${param}`,
+      getTree: param => `${param}`
+    };
+
+    const computedGetters = makeComputedGetters(mockGetters);
+
+    expect(computedGetters.getProducts('test')).toEqual(ref('test'));
+    expect(computedGetters.getTree('test')).toEqual(ref('test'));
```
patzick: shouldn't be read-only computed?

packages/core/interfaces/src/index.ts
```diff
@@ -35,11 +37,8 @@ export interface UseUser
     username: string;
     password: string;
     [x: string]: any;
-  }) => Promise<void>;
-  logout: () => Promise<void>;
-  changePassword: (
-    currentPassword: string,
-    newPassword: string
+  }) => Promise<void>; logout: () => Promise<void>; changePassword: (currentPassword: string,
```
patzick: please fix this formatting, there is something wrong here

packages/core/interfaces/src/index.ts
patzick: this is important!

packages/core/interfaces/src/index.ts
patzick: +1
PRODUCT_TYPES enum + getType sounds great!

packages/core/factories/src/useCartFactory.ts
```diff
@@ -116,6 +117,7 @@ export function useCartFactory<CART, CART_ITEM, PRODUCT, COUPON>(
 
     return {
       cart: computed(() => factoryParams.cart.value),
+      cartGetters: factoryParams.cartGetters,
```
patzick, line 120: makeComputedGetters should be used here

packages/core/factories/src/useCategoryFactory.ts
```diff
@@ -25,6 +26,7 @@ export function useCategoryFactory<CATEGORY, CATEGORY_SEARCH_PARAMS>(
 
     return {
       search,
+      categoryGetters: factoryParams.categoryGetters,
```
patzick, line 29: makeComputedGetters should be used here

packages/core/factories/src/useProductFactory.ts
```diff
@@ -38,10 +39,11 @@ export function useProductFactory<PRODUCT, PRODUCT_SEARCH_PARAMS>(
     };
 
     return {
-      products: computed(() => products.value),
-      totalProducts: computed(() => totalProducts.value),
       search,
-      loading: computed(() => loading.value)
+      productGetters: factoryParams.productGetters,
```
patzick, line 43: makeComputedGetters should be used here

patzick (changes requested):
few notes, yet `packages/core/interfaces/src/index.ts` Getters interfaces need to be discussed with @filrak and @andrzejewsky.

Really great job so far! Love `makeComputedGetters`! :)

packages/boilerplate/composables/src/getters/cartHelpers.ts
krskibin: Outside the scope of the task

packages/core/factories/src/useCartFactory.ts
krskibin: To discuss in another task

packages/core/utils/src/makeComputedGetters/index.ts
krskibin: Changed to `ComputedProperty<any>`

packages/core/utils/__tests__/wrap.spec.ts
krskibin: Fixed!

packages/core/utils/__tests__/makeComputedGetters.spec.ts
krskibin: For testing proposes, I used ref to check wheater the computed properties returned the reactive value.

packages/core/interfaces/src/index.ts
krskibin: fixed

packages/core/factories/src/useProductFactory.ts
krskibin: then the factory would need a type directly from integration, wouldn't it be better to send ready computed getters?

andrzejewsky (changes requested):
first of all, test coverage dropped by -4%, too much, that means you have to ignore some files (if it doesn't make sense to count coverage for them) or just write few lines of tests, thanks!

packages/commercetools/composables/src/useProduct/index.ts
```diff
@@ -14,4 +16,7 @@ const productsSearch = async (params: ProductSearch): Promise<SearchResult<Produ
   };
 };
 
-export default useProductFactory<ProductVariant, ProductSearch>({ productsSearch });
+export default useProductFactory<ProductVariant, ProductSearch>({
+  productsSearch,
+  productGetters: makeComputedGetters(productGetters)
```
patzick, line 21: it shouldn't be used here but in factories under the hood 
image

packages/core/factories/src/useProductFactory.ts
patzick: the idea of factory is to take care of repeatable logic and not force integrations to do it on their own :)

packages/core/utils/src/makeComputedGetters/index.ts
```diff
@@ -1,10 +1,11 @@
-import { computed, Ref } from '@vue/composition-api';
-import wrap from '../wrap';
+import { computed } from '@vue/composition-api';
+import { ComputedProperty } from '@vue-storefront/interfaces';
+import { wrap } from '../wrap';
 
-interface Getters {
-  [key: string]: (param: unknown, ...args: any) => Readonly<Ref<Readonly<unknown>>>;
+export interface Getters {
```
patzick, line 5:
```suggestion
export interface ComputedGetters {
```

patzick (changes requested):
just moving `makeComputedGetters` into factories + change core/interfaces for getters, which was sent on slack

andrzejewsky:
closed here: https://github.com/DivanteLtd/next/pull/297

PR #234, opened 2020-03-05
Add auth middleware for My Account pages

### LINKED ISSUE #216: Limit access to my account pages for logged user only

Qrzy:
- auth middleware added to core theme using `useUser` composable
- changed behavior of `account` icon in app header

Closes #216


packages/commercetools/theme/plugins/api-client.js
```diff
@@ -0,0 +1,24 @@
...
+   * Setup Shopware API client
+   */
+  setup({
+    ...config,
+    contextToken
+  });
+
+  onTokenChange(token => {
+    try {
+      app.$cookies.set(config.cookies.tokenCookieName, token, { maxAge: 60 * 60 * 24 * 365 });
```
defudef: What kind of exception it can throw?

packages/core/theme-module/theme/middleware/auth.js
```diff
@@ -0,0 +1,10 @@
+import { useUser } from '@vue-storefront/commercetools-composables';
+
+const auth = ({ redirect }) => {
+  const { isAuthenticated } = useUser();
+  if (isAuthenticated && !isAuthenticated.value) {
+    return redirect('/');
+  }
+};
+
+export default auth;
```
defudef: what about exporting function itself?
```TS
export default ({ redirect }) => {...}
```

packages/commercetools/api-client/src/index.ts
```diff
@@ -53,17 +53,26 @@ const setup = <TCacheShape>(setupConfig: SetupConfig<TCacheShape>): ApolloClient
   return apolloClient;
 };
 
+const tokenChangeListeners = [];
+
+const onTokenChange = (fn: (token) => void): void => {
+  tokenChangeListeners.push(fn);
+};
+
+const tokenChanged = (newToken) => tokenChangeListeners.forEach(fn => fn(newToken));
```
defudef: IMO these functions are named actually the same and it's not readable because of that.
What do you think to rename the second one into: `changeToken(newToken)` ?

Also could you add missing types? (newToken I guess is string)

defudef (approved):

packages/commercetools/api-client/tests/setup.ts
```diff
@@ -14,6 +14,7 @@ setup({
   cookies: {
     currencyCookieName: 'test-vsf-currency',
     countryCookieName: 'test-vsf-country',
-    localeCookieName: 'test-vsf-locale'
+    localeCookieName: 'test-vsf-locale',
+    tokenCookieName: 'test-vsf-token'
```
filrak: this should be called `authTokenCookieName` it's not descriptive enough

packages/commercetools/theme/nuxt.config.js
```diff
@@ -1,7 +1,7 @@
 import webpack from 'webpack';
-import { config } from './plugins/commercetools.js';
+import { config } from './config.js';
```
filrak: why you moved commercetools config to the root?

packages/commercetools/theme/nuxt.config.js
```diff
@@ -39,11 +39,14 @@ export default {
       });
     }
   },
+  plugins: [
+    './plugins/api-client.js',
```
filrak: why you rename commercetools plugin?

packages/commercetools/theme/plugins/api-client.js
```diff
@@ -0,0 +1,25 @@
...
+import { config } from '../config';
+
+export default ({ app }) => {
+  if (!app.$cookies) {
+    throw 'Error cookie-universal-nuxt module is not applied in nuxt.config.js';
+  }
+  const contextToken = app.$cookies.get(config.cookies.tokenCookieName) || '';
+
+  /**
+   * Setup Shopware API client
```
filrak: ??

packages/core/theme-module/index.js
```diff
@@ -32,14 +30,12 @@ module.exports = function DefaultThemeModule(moduleOptions) {
...
-      assets: '.nuxt/assets',
-      pages: '.nuxt/pages'
-    }};
+    layouts: '.nuxt/layouts',
+    assets: '.nuxt/assets',
+    pages: '.nuxt/pages'
+  };
 
   this.extendBuild(config => {
-    delete config.resolve.alias['~'];
```
filrak, line 42: i think this will break assets resolution

packages/core/theme-module/theme/middleware/auth.js
```diff
@@ -0,0 +1,8 @@
+import { useUser } from '@vue-storefront/commercetools-composables';
```
filrak: path should be taken from config

filrak (changes requested):
If you made this feature custom please write proper documentation (docs package in core).

Also authentication should work by itself without requireing additional code to work. Additional logic (like middleware) should lie automatically added via nuxt module (and have option under config to disable it)

packages/commercetools/theme/nuxt.config.js
Qrzy: I'm not sure, to be honest... I'll move it to `plugins`

packages/commercetools/theme/nuxt.config.js
Qrzy: Back then I thought it was better name since it would match the part it integrates (api client) and since it's in specific integration package there is no need to name it with integration. 
Now, thanks to your comment, I realized it's being copied to destination theme in project so it's really best to keep it named according to the integration. :)

packages/core/theme-module/index.js
Qrzy: Are you sure? 3 lines later it's overwritten anyway, so I think removing this line will not break anything. :)

filrak (approved):

PR #235, opened 2020-03-05
Feature/api docs module

I2olanD:


packages/commercetools/api-client/docs/docs.api.md
```diff
@@ -0,0 +1,130 @@
+## API Report File for "@vue-storefront/docs"
```
filrak: i think it shoudl be called api-client, not docs

packages/core/docs/scripts/index.js
```diff
@@ -0,0 +1,87 @@
+const path = require('path');
```
filrak: i think those scripts should be outside of "docs" folder. Perhaps in the root/scripts as its something that runs for every package but is not a package by itself.

packages/commercetools/api-client/docs/docs.api.md
I2olanD: Actually i have no idea how to change this, configuring the api-extractor...

PR #241, opened 2020-03-09
Feature/about you api (draft)

I2olanD:


packages/about-you/api-client/src/getCategory.ts
```diff
@@ -1,10 +1,11 @@
 
 import { api } from './index';
 import { CategoryWith } from '@aboutyou/backbone/types/CategoryWith';
+import { BapiCategory } from '@aboutyou/backbone/types/BapiCategory';
 
-export type getCategorySearchParams = { ids: number[]; path: string; with: CategoryWith; depth: number} | any
+export type getCategorySearchParams = { ids: number[]; path: string; with: CategoryWith; depth: number}
 
-export default async function(options: getCategorySearchParams) {
+export default async function(options: getCategorySearchParams): Promise<BapiCategory[]> {
```
filrak, line 8: you can set the default param values like that
```suggestion
export default async function(options: getCategorySearchParams = {}): Promise<BapiCategory[]> {
```

packages/about-you/api-client/src/getCategory.ts
```diff
@@ -14,10 +15,11 @@ export default async function(options: getCategorySearchParams) {
       { with: options.with }
     );
   } else if (options.path) {
-    return api.ctagories.getByPath(
+    const response = await api.ctagories.getByPath(
```
filrak:
```suggestion
    const response = await api.categories.getByPath(
```

packages/about-you/api-client/src/getCategory.ts
I2olanD: But this won't work since {} does not fit type getCategorySearchParams. We have to do something like `export default async function(options: getCategorySearchParams = { ids: [], path: '', with: {}, depth: 0, includeHidden: false }):`

I2olanD:
moved to https://github.com/DivanteLtd/next/pull/245

PR #246, opened 2020-03-12
refactor(core):  add common core tests config and useCartFactory tests

patzick:
- [x] common jest config for all core packages
- [x] preparation for common tsconfig file
- [x] common tests coverage for core packages
- [x] useCartFactory tests as an example of test structure
- [x] moved all core tests files to __tests__ directory

TODO in next PRs:
- add tests to theme-module core packages
- refactor useCartFactory -> coupon will not remain after page loading, needs to be provided with a factory method, which would take it from cart


patzick:
Core packages covered, next step -> CT

image

packages/core/utils/src/ssr/index.ts
```diff
@@ -2,9 +2,9 @@ import { getCurrentInstance, onServerPrefetch } from '@vue/composition-api';
 
 type ResourceFunction<T> = (params: any) => Promise<T>
 
-const getRootState = (vm) => {
+const getRootState = (vm: any) => {
   if (vm.$isServer) {
-    return vm.$ssrContext.nuxt.vsfState || {};
```
filrak, line 7: porque?

packages/tsconfig.base.json
```diff
@@ -14,6 +14,9 @@
     "baseUrl": "./",
     "lib": ["es6", "es7"],
     "strict": false,
+    "paths": {
```
filrak, line 17: will be merged anyway

filrak (approved):

packages/tsconfig.base.json
patzick: yup, but there were some problems with recognition from tests point of view as well, it's good to have clear instruction where to search for it for other packages :)

packages/core/utils/src/ssr/index.ts
patzick: line 19 covers this case. So it was not possible to have this alternative executed :)

PR #265, opened 2020-03-17
change interfaces to comply with paginated search results

Qrzy:
Search results in `useProduct` and `useUserAddress` can be paginated, hence this change that follows the way the `useUserOrders` has been prepared for pagination.


packages/core/interfaces/src/index.ts
```diff
@@ -3,8 +3,10 @@ import { Ref } from '@vue/composition-api';
 type ComputedProperty<T> = Readonly<Ref<Readonly<T>>>;
 
 export interface UseProduct<PRODUCT> {
-  products: ComputedProperty<PRODUCT[]>;
-  totalProducts: ComputedProperty<number>;
+  products: {
```
andrzejewsky: please use `products` and `totalProducts` instead

packages/core/interfaces/src/index.ts
```diff
@@ -56,30 +58,27 @@ export interface UseUserOrders<ORDER> {
...
     [x: string]: any;
   }) => Promise<void>;
-  loading: Ref<boolean>;
+  loading: ComputedProperty<boolean>;
 }
 
-/** check if we always have those addresses together or we need pagination  */
 export interface UseUserAddress<ADDRESS> {
-  addresses: ComputedProperty<ADDRESS[]>;
+  addresses: {
```
andrzejewsky: same here `addressess` and` totalAddressess`

andrzejewsky (changes requested):

packages/core/factories/src/useProductFactory.ts
```diff
@@ -9,8 +9,13 @@ type SearchParams = {
   filters?: any;
 }
 
+export type ProductsSearchResult<PRODUCT> = {
+  data: PRODUCT[];
+  total: number;
+};
+
```
patzick:
```suggestion
export type SearchResult<T> = {
  data: T[];
  total: number;
};

```

patzick (changes requested):
Just one note + point from @andrzejewsky

packages/core/factories/src/useProductFactory.ts
patzick: this can be used for all searches

packages/core/factories/src/index.ts
```diff
@@ -1,3 +1,7 @@
+import Vue from 'vue';
+import CompositionApi from '@vue/composition-api';
+Vue.use(CompositionApi);
+
```
patzick:
```suggestion
```

this shouldn't be here

patzick (changes requested):
Just one thing from me :)

patzick (approved):
👍

andrzejewsky (approved):

PR #268, opened 2020-03-19
fix(core-theme): remove zero badge when no items in cart

Qrzy:
removed the _zero_ bagde from cart icon when there are no items in cart


packages/core/theme-module/theme/components/AppHeader.vue
```diff
@@ -50,7 +50,8 @@ export default {
     };
     const { cart } = useCart();
     const cartTotalItems = computed(() => {
-      return getCartTotalItems(cart.value).toString();
```
filrak, line 53: can't we handle it on UI?

Qrzy: We can, but don't you think that even such micro-logic (if zero, make it null) doesn't fit in template?

Qrzy: Issue reported to Storefront UI team: https://github.com/DivanteLtd/storefront-ui/issues/922

defudef (approved):

PR #271, opened 2020-03-19
Use user factory

### LINKED ISSUE #240: Extract useUser to factory
Similarly to what we've done here with to other composables https://github.com/DivanteLtd/next/pull/229 extract reusable logic from `useUser` composable to `useUserFactory`.

Keep in mind that
- platform-specific part should be as small as possible
- if something could work differently on other platform it should also be extracted to factory parameter

Propose PR to `new-core-interfaces`

TODO:

- [ ] extract `useuser` agnostic logic to `useUserFactory`
- [ ] update `useUser` composable in commercetools integration
- [ ] extract agnostic tests to `factories` package
- [ ] make sure tests for your composable are passing

pr0xyMity:


packages/core/factories/__tests__/useUserFactory.spec.ts
```diff
@@ -0,0 +1,121 @@
...
+  logOut: jest.fn(),
+  updateUser: jest.fn(),
+  register: jest.fn(),
+  logIn: jest.fn(),
+  changePassword: jest.fn()
+} as any;
+
+const useUserMethods = useUserFactory(factoryParams)();
+const consoleErrorSpy = jest.spyOn(console, 'error');
+describe('useUserFactory', () => {
```
patzick: please take a look at describes pattern in factories, for example here should be
```suggestion
describe('[CORE - factories] useUserFactory', () => {
```
but go thru others as well. Good describing and naming tests is a great documentation later

packages/core/factories/src/useUserFactory.ts
```diff
@@ -0,0 +1,113 @@
...
+  updateUser: (params: {currentUser: USER; paramsToUpdate: UPDATE_USER_PARAMS}) => Promise<USER>;
+  register: (params: { email: string; password: string; firstName?: string; lastName?: string }) => Promise<USER>;
+  logIn: (params: { username: string; password: string }) => Promise<({updatedUser: USER; updatedCart: CART})>;
+  changePassword: (params: {currentUser: USER; currentPassword: string; newPassword: string}) => Promise<USER>;
+};
+
+export function useUserFactory<USER, CART, UPDATE_USER_PARAMS>(
+  factoryParams: UseUserFactoryParams<USER, CART, UPDATE_USER_PARAMS>
+) {
+  const user: Ref<USER> = ref({});
```
patzick: i think should be
```suggestion
  const user: Ref<USER> = ref(null);
```

packages/commercetools/composables/src/useUser/authenticate.ts
```diff
@@ -0,0 +1,15 @@
+import {CustomerSignMeInDraft, CustomerSignMeUpDraft} from '@vue-storefront/commercetools-api/lib/src/types/GraphQL';
```
filrak, line 1: is there a reason for it to be in different file than factory params?
It's a part of that module

packages/core/factories/src/useUserFactory.ts
```diff
@@ -0,0 +1,113 @@
+import { ref, Ref, watch, computed } from '@vue/composition-api';
+import { UseUser } from '@vue-storefront/interfaces';
+
+export type UseUserFactoryParams<USER, CART, UPDATE_USER_PARAMS> = {
+  cart: Ref<CART>;
+  loadUser: () => Promise<USER>;
+  logOut: (params?: {currentUser?: USER}) => Promise<void>;
+  updateUser: (params: {currentUser: USER; paramsToUpdate: UPDATE_USER_PARAMS}) => Promise<USER>;
```
filrak: shouldn't that be always an updated `currentUser` ? is there a use case when we selectively send only a subset of form elements?

packages/core/factories/src/useUserFactory.ts
```diff
@@ -0,0 +1,113 @@
...
+      }
+      loading.value = true;
+      try {
+        user.value = await factoryParams.loadUser();
+        loading.value = false;
+      } catch (err) {
+        console.log('useUser:Factory:Watch', err);
+        loading.value = false;
+      }
+    });
```
patzick: this whole thing can be replaced same as `refreshCart` in cart factory, see how it is invoked before return section there :)
We should avoid watches like this

packages/commercetools/composables/src/useUser/authenticate.ts
pr0xyMity: yes, I've spoke with @lukeromanowicz about it and it is easier to test it alone and it's cleaner, like in the useCart composable

packages/core/factories/src/useUserFactory.ts
```diff
@@ -0,0 +1,113 @@
...
+
+    const updateUser = async (params: UPDATE_USER_PARAMS) => {
+      loading.value = true;
+      try {
+        user.value = await factoryParams.updateUser({currentUser: user.value, paramsToUpdate: params});
+        loading.value = false;
+      } catch (err) {
+        console.error('useUser:Factory:updateUser', err);
+        loading.value = false;
```
patzick, line 30: loading=false should go to `finally` block
also, catch should rethrow error, because otherwise, the user doesn't know that something went wrong and cant react on that

packages/core/factories/src/useUserFactory.ts
```diff
@@ -0,0 +1,113 @@
...
+      firstName?: string;
+      lastName?: string;
+    }) => {
+      loading.value = true;
+      try {
+        user.value = await factoryParams.register(registerUserData);
+        loading.value = false;
+      } catch (err) {
+        console.error('useUser:Factory:register', err);
+        loading.value = false;
```
patzick, line 41: same as updateUser -> and all methods with loading and catch

packages/core/factories/src/useUserFactory.ts
pr0xyMity: if I understand correctly, you want to keep it always like that?
```suggestion
updateUser: (params: {currentUser?: USER; paramsToUpdate: UPDATE_USER_PARAMS}) => Promise<USER>;
```

packages/core/factories/src/useUserFactory.ts
```diff
@@ -0,0 +1,113 @@
...
+
+    const login = async (loginUserData: {
+      username: string;
+      password: string;
+    }) => {
+      loading.value = true;
+      try {
+        const { updatedUser, updatedCart } = await factoryParams.logIn(loginUserData);
+        user.value = updatedUser;
+        factoryParams.cart.value = updatedCart;
```
patzick: user shouldn't manage the cart, we have useCart for that. Here we could just invoke `refreshCart` from other composable

packages/commercetools/composables/src/useUser/authenticate.ts
filrak: It's argueable if its cleaner but I get the argument with testing

packages/core/factories/src/useUserFactory.ts
```diff
@@ -0,0 +1,113 @@
+import { ref, Ref, watch, computed } from '@vue/composition-api';
+import { UseUser } from '@vue-storefront/interfaces';
+
+export type UseUserFactoryParams<USER, CART, UPDATE_USER_PARAMS> = {
+  cart: Ref<CART>;
+  loadUser: () => Promise<USER>;
+  logOut: (params?: {currentUser?: USER}) => Promise<void>;
+  updateUser: (params: {currentUser: USER; paramsToUpdate: UPDATE_USER_PARAMS}) => Promise<USER>;
+  register: (params: { email: string; password: string; firstName?: string; lastName?: string }) => Promise<USER>;
+  logIn: (params: { username: string; password: string }) => Promise<({updatedUser: USER; updatedCart: CART})>;
```
patzick: this is useCart job to handle cart
```suggestion
  logIn: (params: { username: string; password: string }) => Promise<({updatedUser: USER})>;
```

packages/core/factories/src/useUserFactory.ts
filrak: not  entirely. What i meant is `paramsToUpdate` is implying that we are passing only a subset of `currentUser` data that we want to update while in most cases when you save a form you're sending all the values, not only the updated ones therefore it should be something like `updatedUserData` or similar

patzick (changes requested):
Thanks, there are a couple of things here from me. For sure we need tests description consistency here :) also
* please replace all @ts-ignore comments and handle them properly with typescript
* methods should rethrow exceptions and needs finally block for loading flag
* watch shouldn't be necessary for factory
* cart elements should be handled by cart composable

packages/core/factories/src/useUserFactory.ts
patzick: I believe what Filip has in mind:
```suggestion
  updateUser: (params: {currentUser: USER}) => Promise<USER>;
```

pr0xyMity: Now it's clear thanks 👌🏻

filrak: no it's not @patzick  :D

packages/core/factories/src/useUserFactory.ts
pr0xyMity: should I add something similar to refreshCart in useUser like refreshUser?

packages/core/factories/src/useUserFactory.ts
pr0xyMity:
```suggestion
  updateUser: (params: {currentUser: USER; updatedUserData: UPDATE_USER_PARAMS}) => Promise<USER>;
```

pr0xyMity: Done

packages/core/factories/__tests__/useUserFactory.spec.ts
pr0xyMity: Done

packages/core/factories/src/useUserFactory.ts
pr0xyMity: Done

packages/core/factories/src/useUserFactory.ts
pr0xyMity: Done

packages/core/factories/src/useUserFactory.ts
pr0xyMity: Done

packages/core/factories/src/useUserFactory.ts
pr0xyMity: Done

packages/core/factories/src/useUserFactory.ts
pr0xyMity: Done

packages/core/factories/src/useUserFactory.ts
pr0xyMity: Done

packages/core/factories/src/useUserFactory.ts
```diff
@@ -0,0 +1,104 @@
...
+        loading.value = false;
+      }
+    };
+
+    const refreshUser = async () => {
+      try {
+        user.value = await factoryParams.loadUser();
+      } catch (err) {
+        throw new Error(err);
+      }
```
patzick:
```suggestion
      loading.value = true;
      try {
        user.value = await factoryParams.loadUser();
      } catch (err) {
        throw new Error(err);
      } finally {
        loading.value = false;
      }
```

packages/core/factories/src/useUserFactory.ts
```diff
@@ -0,0 +1,104 @@
...
+        loading.value = false;
+      }
+    };
+
+    const register = async (registerUserData: {
+      email: string;
+      password: string;
+      firstName?: string;
+      lastName?: string;
+    }) => {
```
patzick: needs to be more generic
```suggestion
    const register = async (registerUserData: REGISTER_USER_PARAMS) => {
```

packages/core/factories/src/useUserFactory.ts
```diff
@@ -0,0 +1,104 @@
...
+export type UseUserFactoryParams<USER, UPDATE_USER_PARAMS> = {
+  loadUser: () => Promise<USER>;
+  logOut: (params?: {currentUser?: USER}) => Promise<void>;
+  updateUser: (params: {currentUser: USER; updatedUserData: UPDATE_USER_PARAMS}) => Promise<USER>;
+  register: (params: { email: string; password: string; firstName?: string; lastName?: string }) => Promise<USER>;
+  logIn: (params: { username: string; password: string }) => Promise<USER>;
+  changePassword: (params: {currentUser: USER; currentPassword: string; newPassword: string}) => Promise<USER>;
+};
+
+export function useUserFactory<USER, UPDATE_USER_PARAMS>(
```
patzick: needed for registration
```suggestion
export function useUserFactory<USER, UPDATE_USER_PARAMS, REGISTER_USER_PARAMS >(
```

packages/core/factories/src/useUserFactory.ts
```diff
@@ -0,0 +1,104 @@
...
+import { UseUser } from '@vue-storefront/interfaces';
+
+export type UseUserFactoryParams<USER, UPDATE_USER_PARAMS> = {
+  loadUser: () => Promise<USER>;
+  logOut: (params?: {currentUser?: USER}) => Promise<void>;
+  updateUser: (params: {currentUser: USER; updatedUserData: UPDATE_USER_PARAMS}) => Promise<USER>;
+  register: (params: { email: string; password: string; firstName?: string; lastName?: string }) => Promise<USER>;
+  logIn: (params: { username: string; password: string }) => Promise<USER>;
+  changePassword: (params: {currentUser: USER; currentPassword: string; newPassword: string}) => Promise<USER>;
+};
```
patzick:
```suggestion
export type UseUserFactoryParams<USER, UPDATE_USER_PARAMS, REGISTER_USER_PARAMS> = {
  loadUser: () => Promise<USER>;
  logOut: (params?: {currentUser?: USER}) => Promise<void>;
  updateUser: (params: {currentUser: USER; updatedUserData: UPDATE_USER_PARAMS}) => Promise<USER>;
  register: (params: REGISTER_USER_PARAMS) => Promise<USER>;
  logIn: (params: { username: string; password: string }) => Promise<USER>;
  changePassword: (params: {currentUser: USER; currentPassword: string; newPassword: string}) => Promise<USER>;
};
```

patzick (changes requested):
Looks good, just small things to change. The most important is registration - every integration can have different registration params so needs to be generic :)

packages/core/factories/src/useUserFactory.ts
pr0xyMity: Done

packages/core/factories/src/useUserFactory.ts
pr0xyMity: Done

packages/core/factories/src/useUserFactory.ts
pr0xyMity: Done

packages/core/factories/src/useUserFactory.ts
pr0xyMity: Done

andrzejewsky (approved):

packages/core/factories/src/useUserFactory.ts
```diff
@@ -0,0 +1,105 @@
...
+import { UseUser } from '@vue-storefront/interfaces';
+
+export type UseUserFactoryParams<USER, UPDATE_USER_PARAMS, REGISTER_USER_PARAMS> = {
+  loadUser: () => Promise<USER>;
+  logOut: (params?: {currentUser?: USER}) => Promise<void>;
+  updateUser: (params: {currentUser: USER; updatedUserData: UPDATE_USER_PARAMS}) => Promise<USER>;
+  register: (params: REGISTER_USER_PARAMS) => Promise<USER>;
+  logIn: (params: { username: string; password: string }) => Promise<USER>;
+  changePassword: (params: {currentUser: USER; currentPassword: string; newPassword: string}) => Promise<USER>;
+  refreshUser: () => Promise<void>;
```
patzick: it's not the factory param, it's UseUser method, which is invoking loadUser :)
```suggestion
```

patzick (changes requested):
Just one thing and we can merge that :)

patzick (approved):
From me "okejka", if only checks will pass :)

PR #272, opened 2020-03-19
Fix loading catalog issues

andrzejewsky:


packages/core/interfaces/src/index.ts
```diff
@@ -1,3 +1,5 @@
+/* istanbul ignore file */
```
patzick: I believe you could add in `collectCoverageFrom` entry:
```
'!interfaces/**/*',
```
as coverage from interfaces should not be collected at all

Qrzy (approved):

patzick (changes requested):
just one thing from me for now

filrak (approved):

PR #277, opened 2020-03-19
Fix SSR issues, new API

andrzejewsky:


packages/core/utils/src/event-bus/index.ts
```diff
@@ -0,0 +1,47 @@
+const listeners: Listener[] = [];
+let emitted: Event[] = [];
+
+type Handler = (value?: any) => void;
```
patzick: as this is exposed as external param, should be exported as well
```suggestion
export type Handler = (value?: unknown) => void;
```

packages/core/utils/src/event-bus/index.ts
```diff
@@ -0,0 +1,47 @@
...
+
+interface Listener {
+  eventName: string;
+  handler: Handler;
+}
+
+interface EventBus {
+  emit: (eventName: string, value: any) => void;
+  on: (eventName: string, handler: Handler) => void;
+}
```
patzick: I believe we don't need this one, methods exports same interface
```suggestion
```

packages/core/utils/src/event-bus/index.ts
```diff
@@ -0,0 +1,47 @@
...
+
+const emit = (eventName: string, value: any): void => {
```
patzick:
```suggestion
export function emit (eventName: string, value: any): void {
```

packages/core/utils/src/event-bus/index.ts
```diff
@@ -0,0 +1,47 @@
+const listeners: Listener[] = [];
+let emitted: Event[] = [];
```
patzick: we should move it under types declaration, but before exported functions

packages/core/utils/src/event-bus/index.ts
```diff
@@ -0,0 +1,47 @@
...
+  relatedListeners.forEach((listener) => {
+    listener.handler(value);
+  });
+
+  if (relatedListeners.length === 0) {
+    emitted.push({ eventName, value });
+  }
+};
+
+const on = (eventName: string, handler: Handler): void => {
```
patzick:
```suggestion
export function on (eventName: string, handler: Handler): void {
```

packages/core/utils/src/event-bus/index.ts
```diff
@@ -0,0 +1,47 @@
...
+  notConsumed.forEach((event) => {
+    handler(event.value);
+  });
+
+  emitted = emitted.filter(event => event.eventName !== eventName);
+};
+
+const eventBus: EventBus = { emit, on };
+
+export default eventBus;
```
patzick: we're exporting methods directly
```suggestion
```

packages/core/factories/src/useCategoryFactory.ts
```diff
@@ -10,16 +10,16 @@ export function useCategoryFactory<CATEGORY, CATEGORY_SEARCH_PARAMS>(
   factoryParams: UseCategoryFactoryParams<CATEGORY, CATEGORY_SEARCH_PARAMS>
 ) {
   return function useCategory(cacheId?: string): UseCategory<CATEGORY> {
-    const { state, persistedResource } = usePersistedState(cacheId);
+    const { state } = useSSR(cacheId);
```
patzick: would be nice to have instead
```
const { state, setCache } = useSSR(cacheId) 
```
then it's just call
```
setCache(categories.value)
```
and even cacheId is set at this point internally

packages/core/factories/src/useCategoryFactory.ts
```diff
@@ -10,16 +10,16 @@ export function useCategoryFactory<CATEGORY, CATEGORY_SEARCH_PARAMS>(
...
-      loading.value = true;
-      categories.value = await persistedResource<CATEGORY[]>(
-        factoryParams.categorySearch,
-        params
-      );
+      if (!state) {
+        loading.value = true;
+      }
+      categories.value = await factoryParams.categorySearch(params);
+      eventBus.emit('set-ssr-cache', { key: cacheId, value: categories.value });
```
patzick: after other comment
```suggestion
      setCache(categories.value)
```
wdyt @andrzejewsky @filrak ?

packages/core/theme-module/theme/pages/Category.vue
```diff
@@ -253,26 +255,32 @@ import {
   getCategoryTree,
   getProductVariants
 } from '<%= options.helpers %>';
+import { onSSR } from '@vue-storefront/utils';
 
 export default {
   transition: 'fade',
+  asyncData() {
```
filrak: wtf?

filrak (changes requested):

packages/core/theme-module/theme/pages/Category.vue
andrzejewsky: @filrak ić stont, WIP!!

packages/core/factories/src/useCategoryFactory.ts
```diff
@@ -10,16 +10,16 @@ export function useCategoryFactory<CATEGORY, CATEGORY_SEARCH_PARAMS>(
...
 
     const search = async (params: CATEGORY_SEARCH_PARAMS) => {
-      loading.value = true;
-      categories.value = await persistedResource<CATEGORY[]>(
-        factoryParams.categorySearch,
-        params
-      );
+      if (!state) {
+        loading.value = true;
+      }
```
patzick: this is not necessary :) 
```suggestion
      loading.value = true;
```

packages/core/factories/src/useProductFactory.ts
```diff
@@ -17,23 +18,17 @@ export function useProductFactory<PRODUCT, PRODUCT_SEARCH_PARAMS>(
...
-    // const products: Ref<ProductVariant[]> = ref(state || []);\
-    const products: Ref<PRODUCT[]> = ref([]);
-    const totalProducts: Ref<number> = ref(0);
+    const { state, saveCache } = useSSR(cacheId);
+    const products: Ref<PRODUCT[]> = ref(state || []);
+    const totalProducts = ref(0);
     const loading = ref(false);
 
-    const search = async (params: PRODUCT_SEARCH_PARAMS) => {
+    const search = async (params: any) => {
```
patzick: why is that?
we need to use `PRODUCT_SEARCH_PARAMS` here :)
```suggestion
    const search = async (params: PRODUCT_SEARCH_PARAMS) => {
```

patzick (changes requested):
Tho small changes from me, overriding this looks 👌

packages/core/utils/src/event-bus/index.ts
andrzejewsky: noo this is only for internal usage, it's not used anywhere

packages/core/docs/core/utils/server-side-rendering.md
```diff
@@ -0,0 +1,106 @@
...
+
+## Usage
+
+In the most cases, the composable functions provide SSR support (it is always transparent for developer).
+To build that composable you have to create a new composable-factory with using `useSSR` function.
+
+```js
+import { useSSR } from '@vue-storefront/utils';
+
+export function useExampleFactory(factoryParams) {
```
filrak: imo factory here is not needed and just adds redundant complexity, just show how to use it with some example use case (using `example` is not the best way to explain stuff as well, show something that can be understood as an idea - simplified product or so)

filrak (changes requested):
Regarding API ( `saveToCache` seems like the right one and if we're saving to cache then we should retrieve `cache`, not state (or the opposite we should save to `state` not cache)

packages/core/theme-module/theme/components/RelatedProducts.vue
```diff
@@ -40,27 +38,18 @@ export default {
...
-  setup({ product }) {
-    const { products, search, loading } = useProduct('related-products');
-    const categories = getProductCategories(product);
-    const relatedProducts = computed(() => getProductVariants(products.value, { masters: true }).filter((prod) => getProductId(prod) !== getProductId(product)));
-
-    if (categories.length > 0) {
-      search({ catId: [categories[0]] });
-    }
-
+  setup() {
```
filrak: shouldn't setup be always on top?

filrak (approved):

PR #290, opened 2020-03-24
feat(core): Apiclientfactory

filrak:


packages/core/factories/src/apiClientFactory.ts
```diff
@@ -0,0 +1,18 @@
+export function apiClientFactory<API_CLIENT_METHODS, API_CLIENT_SETTINGS>(factoryParams: { defaultSettings; onSetup }) {
+  let settings = { ...factoryParams.defaultSettings };
```
andrzejewsky:
```
let settings = { ...factoryParams.defaultSettings, overrides: {} };
```

andrzejewsky (changes requested):
looks ok 👍 just one comment and look what's wrong with the build

packages/core/factories/src/apiClientFactory.ts
```diff
@@ -0,0 +1,18 @@
...
+    },
+    setup (config: API_CLIENT_SETTINGS) {
+      settings = { ...factoryParams.defaultSettings, overrides: settings.overrides, ...config };
+      factoryParams.onSetup(settings);
+    },
+    update (config: API_CLIENT_SETTINGS) {
+      settings = { ...settings, ...config };
+      factoryParams.onSetup(settings);
+    },
+    getSettings: () => Object.freeze(settings)
```
patzick: this should return type with `API_CLIENT_SETTINGS ` for IDE

patzick (approved):
I'd need to see the use case for this factory, but overall looks okay to me :)

andrzejewsky (approved):

PR #297, opened 2020-03-26
Introduce new helpers interfaces

andrzejewsky:


packages/commercetools/composables/src/getters/cartGetters.ts
```diff
@@ -0,0 +1,73 @@
...
+
+export const getCartItems = (cart: Cart): LineItem[] => {
+  if (!cart) {
+    return [];
+  }
+
+  return cart.lineItems;
+};
+
+export const getCartItemName = (product: LineItem): string => product.name;
```
filrak, line 13: shouldn't it be `cartItem` ? its not always a product, for example in shopware it can be a discount

filrak (changes requested):
Change package name from `commercetools-composables` to `commercetools`

packages/core/interfaces/src/index.ts
```diff
@@ -39,11 +39,10 @@ export interface UseUser
   logout: () => Promise<void>;
   changePassword: (
     currentPassword: string,
-    newPassword: string
-  ) => Promise<void>;
+    newPassword: string) => Promise<void>;
+  refreshUser: () => void;
```
patzick: why it's not promise anymore?
```suggestion
  refreshUser: () => Promise<void>;
```

packages/core/interfaces/src/index.ts
```diff
@@ -159,39 +160,107 @@ export interface UseLocale
...
+  getItems: (cart: CART) => CART_ITEM[];
+  getItemName: (cartItem: CART_ITEM) => string;
+  getItemImage: (cartItem: CART_ITEM) => string;
+  getItemPrice: (cartItem: CART_ITEM) => AgnosticPrice;
+  getItemQty: (cartItem: CART_ITEM) => number;
+  getItemAttributes: (cartItem: CART_ITEM, filters?: Array<string>) => Record<string, AgnosticAttribute | string>;
+  getItemSku: (cartItem: CART_ITEM) => string;
+  getTotals: (cart: CART) => AgnosticTotals;
+  getShippingPrice: (cart: CART) => number;
+  getTotalItems: (cart: CART) => number;
```
patzick, line 188: the same as `getItems.length`, so we don't need that here as getItems always returns an array
```suggestion
```

filrak (approved):

packages/core/interfaces/src/index.ts
```diff
@@ -159,39 +160,107 @@ export interface UseLocale
...
+  [getterName: string]: (element: any, options?: any) => unknown;
+}
+
+export interface UserOrderGetters<ORDER> {
+  getDate: (order: ORDER | ComputedProperty<ORDER>) => string;
+  getId: (order: ORDER | ComputedProperty<ORDER>) => string;
+  getStatus: (order: ORDER | ComputedProperty<ORDER>) => string;
+  getPrice: (order: ORDER | ComputedProperty<ORDER>) => number;
+  [getterName: string]: (element: any, options?: any) => unknown;
+}
```
patzick: we need to remove ComputedProperty param options here

patzick (changes requested):

patzick (approved):

Issue #274, opened 2020-03-19
[RFC] Improve loading state


lukeromanowicz:
## Current state:
all composables have only one loading state shared between all actions which makes it:
1) hard to manage when there is series of actions to perform
2) not too flexible when it comes to creating UI with different types of loading

## Proposed change:
Change single loading property in composables to be a map of loading states for every action.

e.g.
in:
```
  return {
    ...
    availableLocales,
    availableCountries,
    availableCurrencies,
    loading: computed(() => loading.value)
  }
```
loading should become an object containing following bool properties: availableLocales, availableCountries, availableCurrencies.

Optional:
we could also add property: 'any' that would be true if any of sub-loading properties is true, otherwise false.

PR #300, opened 2020-03-26
refactor: implement factory for useLocale

lukeromanowicz:


lukeromanowicz:
I am not 100% sure if useLocale is useful enough. There is "locale" which isn't really defined if it's a language or language+country combined. If it's not language then I have a feeling there should be a separate language field. It also lacks providing available languages based on a selected country, so the filtering logic has to be implemented per platform/shop implementation if needed.

lukeromanowicz:
tests fail due to #301

packages/commercetools/composables/src/types/Locale.ts
```diff
@@ -0,0 +1,9 @@
+import { Ref } from '@vue/composition-api';
+import { LocaleItem } from '@vue-storefront/commercetools-api/lib/src/types/setup';
+
```
filrak: We should pass only data types here and take care of Refs/computed properties typing inside the factory - like in other factories. By giving a freedom of deciding what it is we can lose possibility to modify that by ourselves in the future for all integrations in one place

packages/commercetools/composables/src/types/Locale.ts
```diff
@@ -0,0 +1,9 @@
...
+export type Locale = Ref<string>
+export type Country = Ref<string>
+export type Currency = Ref<string>
+export type AvailableLocales = Ref<Readonly<LocaleItem[]>>
```
filrak: If we apply the above we can also get rid of this type (and `AvailableCurrencies`, `AvailableCountries`

packages/core/factories/src/useLocaleFactory.ts
```diff
@@ -0,0 +1,88 @@
...
+  >(factoryParams: UseLocaleFactoryParams<
+  LOCALE,
+  COUNTRY,
+  CURRENCY,
+  AVAILABLE_LOCALES,
+  AVAILABLE_COUNTRIES,
+  AVAILABLE_CURRENCIES>
+) {
+  let loadingLocale = false;
+  const localeCache: Ref<LOCALE> = ref(null);
```
filrak: can you elaborate why you called that `cache` - it doesn't seem to cache anything, perhaps using underscore to indicate it's a [private property might be better?

filrak (changes requested):
@andrzejewsky can you elaborate on why this `locale` is needed apart from having `country` and `currency` is this kind of a region?

Also this PR breaking localization on Commercetools - did @andrzejewsky  agreed on that?

andrzejewsky:
@filrak `locale` it a selected language (we can change it to the proper name if needed). I know that breaks CT lang selector, but it's a part of the refactoring by using new factories - we have task on our board to provide implementation - @lukeromanowicz creates here just factory

packages/core/factories/src/useLocaleFactory.ts
lukeromanowicz: @filrak I used naming "cache" to indicate that it's storing the state for computed because 'locale' is occupied by the computed itself. I am not really a fan of underscore, especially that I am not aware of using underscore naming anywhere else in this project

packages/commercetools/composables/src/types/Locale.ts
lukeromanowicz: I am using the exact same types that were here before "refactoring". It requires further work from CT team anyway to fully implement it.

lukeromanowicz:
@filrak @andrzejewsky As far as I understand our agreement, I was supposed to do the refactoring for the core part and leave the types the same as they were in CT package, which later will be improved by CT team. Am I wrong about that?

packages/commercetools/composables/src/types/Locale.ts
filrak: Well the task was also to refactor current version of implemented comparable, the job should be done on the factory side because this is where you pass the types ;)

packages/core/factories/src/useLocaleFactory.ts
filrak: Underscore is used in few places like enhancers. Imho its more descriptive than `cache` as it could be misleading @patzick @andrzejewsky wdyt?

packages/core/interfaces/src/index.ts
```diff
@@ -147,16 +147,20 @@ export interface UseLocale
...
-  currency: CURRENCY;
-  availableLocales: AVAILABLE_LOCALES;
-  availableCountries: AVAILABLE_COUNTRIES;
-  availableCurrencies: AVAILABLE_CURRENCIES;
+  locale: ComputedProperty<LOCALE>;
+  country: ComputedProperty<COUNTRY>;
+  currency: ComputedProperty<CURRENCY>;
+  setLocale: (locale: LOCALE) => void;
+  setCountry: (country: COUNTRY) => void;
+  setCurrency: (currency: CURRENCY) => void;
```
patzick: I'd add promises here to cover cases, where integration may store these pieces of information on API
```suggestion
  setLocale: (locale: LOCALE) => Promise<void>;
  setCountry: (country: COUNTRY) => Promise<void>;
  setCurrency: (currency: CURRENCY) => Promise<void>;
```

filrak:
You're right, but the factory types are wrong. Factory should accept only data formats of a given platform without their context (so no info if its a ref or computed property). Please take a look how other factories work and adjust accordingly :)

packages/core/factories/src/useLocaleFactory.ts
```diff
@@ -0,0 +1,88 @@
...
+        (async () => availableLocalesCache.value = await factoryParams.loadAvailableLocales())(),
+        (async () => availableCountriesCache.value = await factoryParams.loadAvailableCountries())(),
+        (async () => availableCurrenciesCache.value = await factoryParams.loadAvailableCurrencies())()
+      ]);
+    } finally {
+      loadingLocale = false;
+    }
+  };
+
+  refreshAvailableElements();
```
patzick: instead of this, add methods to load those lists separately

```suggestion
```

packages/core/factories/src/useLocaleFactory.ts
```diff
@@ -0,0 +1,88 @@
+import {computed, Ref, ref} from '@vue/composition-api';
+import { UseLocale } from '@vue-storefront/interfaces';
+
+export type UseLocaleFactoryParams<
+  LOCALE,
+  COUNTRY,
+  CURRENCY,
+  AVAILABLE_LOCALES = LOCALE[],
+  AVAILABLE_COUNTRIES = COUNTRY[],
+  AVAILABLE_CURRENCIES = CURRENCY[]> = {
```
patzick: we'll use agnostic types for this, so:
* removing additional types here,
* export methods for loading available lists separately
* adding params for loading methods for filters (for example for fetching currencies only for a specific country)

packages/core/factories/src/useLocaleFactory.ts
patzick: currentLocal, currentCountry etc etc would be great :)

patzick (changes requested):
* create agnostic types for Locale, Currency, Country

packages/core/factories/src/useLocaleFactory.ts
lukeromanowicz: the part about loading requires https://github.com/DivanteLtd/next/issues/274

andrzejewsky (approved):

packages/commercetools/api-client/src/index.ts
```diff
@@ -32,7 +32,7 @@ let locales = [];
 let currentToken: Token = null;
 let api: ApiConfig = null;
 let auth: Auth = {
-  onTokenChange: (token: Token) => {},
+  onTokenChange: () => {},
```
filrak, line 35: why

packages/commercetools/composables/src/types/Locale.ts
```diff
@@ -0,0 +1,9 @@
+import { Ref } from '@vue/composition-api';
```
filrak: incude that into main types.ts as well please

packages/core/factories/src/useLocaleFactory.ts
```diff
@@ -0,0 +1,67 @@
+import { computed, Ref, ref } from '@vue/composition-api';
+import { AgnosticCountry, AgnosticCurrency, AgnosticLocale, UseLocale } from '@vue-storefront/interfaces';
+
+export type UseLocaleFactoryParams = {
+  setLocale: (AgnosticLocale) => Promise<void>;
```
filrak: `setX` should always return new state, not void

filrak (changes requested):

packages/commercetools/api-client/src/index.ts
andrzejewsky: eslint

filrak (approved):

PR #325, opened 2020-03-29
Sending CT token via cookies

andrzejewsky:


Qrzy (approved):

filrak (commented):
Update CT API Client docs :)

packages/commercetools/api-client/src/types/setup.ts
```diff
@@ -39,6 +39,9 @@ export interface SetupConfig<TCacheShape> {
   currencies?: LocaleItem[];
   locales?: LocaleItem[];
   cookies?: CookiesConfig;
+  onTokenSave?: (token: Token) => void;
```
filrak: Imho it should be a single object containing those props like auth?

packages/commercetools/api-client/src/helpers/createCommerceToolsLink/tokenCache.ts
```diff
@@ -1,22 +1,16 @@
 import { Token } from './../../types/setup';
```
filrak: Do we really need this file?

filrak (changes requested):

packages/commercetools/api-client/__tests__/helpers/tokenFlow.spec.ts
```diff
@@ -1,30 +1,32 @@
...
-  projectKey: 'project-key',
-  clientId: 'client-id',
-  clientSecret: 'client-secret',
-  scopes: []
+import createAccessToken from './../../src/helpers/createAccessToken';
+import { setup } from './../../src/index';
+
+const config = {
+  api: {
+    uri: 'https://example.com',
```
defudef, line 7: It's a good practice to provide non-existing URLs so nothing weird will go "outside" when things have gone wrong, for example: `https://localhost` or just `https://github.com/DivanteLtd/next`

packages/commercetools/api-client/src/helpers/createAccessToken/index.ts
```diff
@@ -38,24 +39,24 @@ const refreshTokenFlow = async (config: ApiConfig): Promise<Token> => {
 
 const isTokenExpired = (token: Token): boolean => Date.now() > token.expires_at;
 
-const createAccessTokenFlow = async (config: ApiConfig, options: FlowOptions = {}): Promise<Token> => {
+const createAccessToken = async (options: FlowOptions = {}): Promise<Token> => {
```
defudef, line 42: It can return `Promise<undefined>` as well

packages/commercetools/theme/middleware/commercetools.js
```diff
@@ -0,0 +1,16 @@
+import { createAccessToken } from '@vue-storefront/commercetools-api';
+import { setup } from '@vue-storefront/commercetools-api';
+import { config } from './../plugins/commercetools-config';
+
+export default async ({ app }) => {
+  let currentToken = app.$cookies.get('vsf-commercetools-token');
```
defudef: what about to define `vsf-commercetools-token` in config or at least as a constant?

packages/commercetools/theme/middleware/commercetools.js
```diff
@@ -0,0 +1,16 @@
...
+
+  if (!currentToken) {
+    currentToken = await createAccessToken();
+  }
```
defudef, line 12: I think it can be simplified:
```TS
const currentToken = app.$cookies.get('vsf-commercetools-token') || (await createAccessToken());
````

File: packages/commercetools/theme/plugins/commercetools.js [hunk 7]
```diff
@@ -1,10 +1,29 @@
 import { setup } from '@vue-storefront/commercetools-api';
 import { config } from './commercetools-config';
 
-export default () => {
+export default ({ app }) => {
+  const currentToken = app.$cookies.get('vsf-commercetools-token');
```
defudef: Yup,I think constant for a cookie name would be good

defudef (approved):

packages/commercetools/api-client/__tests__/helpers/tokenFlow.spec.ts
filrak: +1, you can also just hardcode our demo instance ;)

packages/commercetools/api-client/src/api/customerSignOut/index.ts
```diff
@@ -1,7 +1,7 @@
-import { cleanToken } from './../../helpers/createCommerceToolsLink/tokenCache';
+import { tokenEvents } from './../../index';
 
 const customerSignOut = async (): Promise<void> => {
-  cleanToken();
+  tokenEvents.onTokenRemove();
```
filrak: i dont think this should be an api client method if its not calling the api

packages/commercetools/api-client/src/index.ts
```diff
@@ -29,33 +31,46 @@ let country = '';
 let countries = [];
 let currencies = [];
 let locales = [];
+let currentToken: Token = null;
+let api: ApiConfig = null;
+let tokenEvents: TokenEvents = {
```
filrak: imho it should be called `auth` so we can group all properties connected to auth here. `tokenEvents` tells absolutely nothing about what these properties are responsible for

packages/commercetools/api-client/src/index.ts
```diff
@@ -29,33 +31,46 @@ let country = '';
...
+  onTokenChange: (token: Token) => {},
```
filrak, line 35: also `on` indicates that something happened before and we are reacting to this event when in reality nothing happens. Shouldn't these methods be called just `changeToken` etc?

filrak (changes requested):
Correct naming please.

Also regarding the documentation you should rather augment the auth middleware document, renami it to auth and describe how the mechanism works, how to override tokens etc.

packages/core/docs/commercetools/api-client.md
```diff
@@ -71,6 +71,50 @@ export interface TokenEvents {
...
+The most common use-case is storing that token using cookies:
+
+```js
+import { setup } from '@vue-storefront/commercetools-api';
+import { config } from './your/config';
+
+const currentToken = app.$cookies.get('ct-token');
+
+const onTokenChange = (token) => {
+  app.$cookies.set('ct-token', token);
```
filrak, line 88: how backend knows that it's a right token, does the cookie always need to be named `ct-token` ?

andrzejewsky: You can name it whatever you like, it's just cookie name that stores the token. Backend doesn't have any validation stuff, if token is invalid the API will return an error, but there is no way to put invalid token there unless you do it manually or there is a bug in the code

packages/commercetools/api-client/__tests__/helpers/tokenFlow.spec.ts
andrzejewsky: it can't go outside, there is a mock underneath

packages/commercetools/api-client/src/helpers/createAccessToken/index.ts
andrzejewsky: but it's a valid typing as well, I don't want to put there Token | null | undefined, because it basically won't change anything

PR #326, opened 2020-03-30
[RFC] Default pagination settings in composables 

Qrzy:
[see rendered](https://github.com/DivanteLtd/next/blob/634478a3d8832633c813a303ea1af3fa9a56874c/packages/core/rfcs/20200330-pagination-default-settings.md)

### DIFF
--- a/packages/core/rfcs/20200330-pagination-default-settings.md
+++ b/packages/core/rfcs/20200330-pagination-default-settings.md
@@ -28,6 +28,9 @@ export type UseProductFactoryParams<PRODUCT, PRODUCT_SEARCH_PARAMS extends Searc
 };
 ```
 
+Why `loadDefaultPaginationSettings` is asynchronous?
+This way it's possible to fetch such settings from the underlying platform.
+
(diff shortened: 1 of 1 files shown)


filrak:
I don't think we need `loadDefaultPaginationSettings` at all. Imho the defaults should be in API Client and if we will override them in search then they are overrided, that's all.

filrak (changes requested):
I don't think we need loadDefaultPaginationSettings at all. Imho the defaults should be in API Client and if we will override them in search then they are overrided, that's all. Also it should be configured in API Client, not compostables - otherwise how you will handle pagination with API Client itself?

Qrzy:
@filrak and what about the case when platform gives possibility to change the default page size on backend? Should we not reflect that immediately?

filrak:
It should be handled on API Client level, directly in getSth method. Pagination is a parameter of search, there is no point in making a separate feature out of it when it isn't a separate feature

filrak:
as discussed - closing

PR #328, opened 2020-03-31
Fix double-path when compiling the templates causing empty templates generated

passtas:
The path is joined already in line 20 and 22 by (dirPath + '/' + file)

yarn build:dev doesn't trigger an error during the build because compileTemplates.js use asynchronous readFile and writeFile, otherwise it would trigger the error like this:

> FATAL  ENOENT: no such file or directory, open 'D:\projects\vue-storefront\packages\core\theme-module\D:\projects\vue-storefront\packages\core\theme-module\theme\assets\css\reset.scss' 

### DIFF
--- a/packages/core/theme-module/index.js
+++ b/packages/core/theme-module/index.js
@@ -34,7 +34,7 @@ module.exports = function DefaultThemeModule(moduleOptions) {
   const compileAgnosticTemplates = () => {
     themeFiles.forEach((file) => {
       compileTemplates(
-        path.join(__dirname, file),
+        file,
         this.options.buildDir.split('.nuxt').pop() + '.theme/' + file.split('theme/').pop(),
         {
           apiClient: moduleOptions.apiClient,


Qrzy:
Hello @passtas 
Thank you for your contribution! However, I think you missed few things here.
On line 20 the function just calls itself recursively for nested directories so it's obviously providing the full path.
But let's take a closer look at line 22:
```javascript
arrayOfFiles.push((dirPath + '/' + file).split(__dirname + '/').pop());
```
the most important part is in the end of line - the path is reduced to the part **after** `__dirname` so there is no doubled path on line 37.

The change introduced in this PR leads to broken build.

filrak:
Ok so turns out using `/` as dir separated was a problem here. Fix incoming, I'm closing this PR then, thanks for participating!

passtas:
Yeah, you are absolutely right `split(__dirname + '/')` won't work for windows because of backslashes are used, so you may need to normalize that first before reducing.

PR #333, opened 2020-04-01
[WIP] refactor: move pagination default settings to api client config

Qrzy:


packages/core/theme-module/theme/pages/Category.vue
```diff
@@ -310,15 +310,11 @@ import {
 } from '@storefront-ui/vue';
 import { computed, ref, watch } from '@vue/composition-api';
 import { useCategory, useProduct, productGetters, categoryGetters } from '<%= options.composables %>';
+import { pagination } from '<%= options.apiClient %>';
```
Qrzy, line 313: @filrak here's the thing we discussed. 
Without exposing those defaults through composable I had to assume api client exposes pagination defaults and use them this way
OR
hardcode it totally, like it was before this PR.
We need to discuss it again. :)
#334

Qrzy:
As discussed, this approach is abandoned.
Defaults will be defined by the user of composable and always passed to search function.

PR #335, opened 2020-04-01
[RFC] Factories composable relations

defudef:

### DIFF
--- a/packages/core/rfcs/composable-relations.md
+++ b/packages/core/rfcs/composable-relations.md
@@ -1,0 +1,86 @@
+# Composable relations
+
+## Motivation
+
+By injecting composables to the others we can achieve an open to extensions dependency management.
+
+## Case study
+
+We have two composables:
+
+- useUser
+- useUserAddress
+
+We use GraphQL CommerceTools and in this case editting user's addresses is through the user collection. This approach require update the user object after each modifications in their adresses. That means we need access to user entity through useUserAddress composable
+
+
+### Current approach
+
+In order to get access to the user through the useUserAddress we need to following changes (please treat this more as a pseudo code, rather than a strict one):
+
+```TS
+// useUser composable
+export const user = ref(null);
+```
+
+```TS
+// useUserAddress composable
+import { user } from 'useUserComposable';
+
+export const params = {
+  addAddress: (address) => {
+    const addresses = await addNewAddressGraphQL(address);
+
+    user.value = { ...user.value, addresses };
+
+    return addresses;
+  }
+};
+```
+
+As we see, this approach has some overcomplexity solutions:
+- mutations in factories and composables as well
+- the user object is now public and anyone can avoid the architecture principles
+- too much logic in composables. The solution is not agnostic anymore
+
+### Proposed approach
+
+Instead of exporting the user object, we can inject an optional (or not) useUser composable through factory params. We can also define logic depending on is useUser provided or not.
+
+Examples:
+
+```TS
+//  useUser composable
+
+const user = ref(null); // user object is now a private property
+```
+
+```TS
+// useUserAddressFactory
+
+interface UseUserAddressFactoryParams {
+  userComposable?: UseUser<USER, USER_UPDATE_PARAMS>;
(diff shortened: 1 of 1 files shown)


packages/core/rfcs/composable-relations.md
```diff
@@ -0,0 +1,86 @@
...
+
+```TS
+// useUserAddressFactory
+
+interface UseUserAddressFactoryParams {
+  userComposable?: UseUser<USER, USER_UPDATE_PARAMS>;
+  addAddress: { userUpdated: USER, addressesUpdated: ADDRESS[] }
+}
+
+export function useUserAddressFactory {
```
andrzejewsky, line 66: that's the weird example it's not even a factory but looks like factory params

filrak:
The intention was very similar. The fact that `useuser` is calling the composable directly is our fault - we missed that during CR somehow.

The only change I would suggest is to provide explicitly a `dependencies` object into params that will be holding all external dependencies. Also i wouldn't suggest passing whole compostable there, rather a specific objects or functions because if we want to modify for example `cart` from `useCart` we can't do that from within comparable. Btw check `useCart` for reference, I suggest something similar so for example in `useUser` we will pass
```js
dependencies: {
    cart: cart
    refreshCart: useCart().refreshCart()
}
```

I really don't get that part - can you elaborate on what was the intention?

```js
export function useUserAddressFactory {
  return {
    userComposable: useUser(),
    addAddress: (address) => {
      const { userUpdated, adressessUpdated } = await factoryParams.addAddress(address);
      addresses.value = addresses;
      userComposable && userComposable.updateUser(userUpdated);
    }
  };
}
```

Imho factory should return only a comparable, whats the point of returning other comparable from this factory along with some methods? I'm suspecting you just made some mistake here and you meant factory params here?

packages/core/rfcs/composable-relations.md
andrzejewsky: I was thinking about this DI etc. and it ONLY make sense when you need to call other composable INSIDE of the factory, but this way is a little bit risky as some of these calls can depend on the integration. Using factory params you can do something like this

```
const someComposable = useSomeComposable();

const factoryParams = {
  someFunction: async () => {
    const field = someComposable.someField;
  }
}
```

andrzejewsky: @defudef so please provide other examples where it won't work and you need DI mechanism

filrak: I'm actually against calling it that way tbh, we should have a way of performing such stuff from the inside of the factory so the control is on our side, O believe this is the only reason why we need such thing. Ideally we should have some kind of internal linking between compostables that needs to be linked (i believe it applies only to user

filrak:
I just came up with another idea, probably much more ellegant. Solution proposed in this RFC is a bit overcomplicated and relies on the factory params in terms of injecting the data - would be nice if we could hide this relation inside factories. 

We can do tha from the inside of the factories. we just need an object that will work like a "mediator" between user cart and user subcomposables.

Mediator object could look like this
```js
const shared = {
  cart: ref(null),
  user: ref(null)
}
```

 Then we're just reading from this mediator object inside compostables that need it. 
```js
import { shared } from './shared'

function useUserAddressFactory() {
  const user = shared.user
  return function useUserAddresses() { ... }
}
```
And overriding if that's needed

```js
import { shared } from './shared'

// from useCartFactory
function useCartFactory () {
  const cart = shared.cart

  // later in code
  loadCart () {
     const updatedCart = factoryParams.loadCart()
     cart.value = updatedCart
  }
```

With that we can omit forcing user to directly inject the properties into factories which brings back the control to us and reduces complexity of the integration.

Another advantage would be keeping full treeshaking - previous approach ends up in grouping shared compostable in the same chunks

packages/core/rfcs/composable-relations.md
defudef: Sorry, just missed brackets:

```TS
export function useUserAddressFactory() {
 // that stuff inside here
}
````

filrak:
Closing. There is a new RFC on that topic https://github.com/DivanteLtd/next/pull/342

PR #341, opened 2020-04-03
[RFC ] Errors handling

andrzejewsky:

### DIFF
--- a/packages/core/rfcs/errors-habdling-in-composables.md
+++ b/packages/core/rfcs/errors-habdling-in-composables.md
@@ -1,0 +1,63 @@
+# Errors handling in composables
+
+## Motivation
+
+We need to handle errors, that are caused by external api calls provided by factory params.
+Furthermore, it must be something to recognize where that error comes from (which call).
+
+We can solve it by introducing fields inside the errors object and the `AgnosticError`:
+
+Factory
+```js
+const useProductFactory = (factoryParams) => {
+  const useProduct = () => {
+    const products = ref([]);
+    const errors = ref({};)
+
+
+    const search = async (params) => {
+      try {
+        products.value = factoryParams.searchProducts(params);
+      } catch (err) {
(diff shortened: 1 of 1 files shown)


filrak (commented):
I like the idea but I don't like the solution. below few concerns from my side;

1. I would suggest to have a default, generic errors in factories like "Something went wrong with searching the products" that can be extended with specific cases for the integration
2. How you want to handle i18n?
3. What if someone wants to change the error message for specific shop (for example they want to add link to support)
4. Wouldn't it be better to just return an `error` object from specific functions in factory params instead of having this `convertError` function? Seems more natural to me to have that in one place and doesn't involve any underlying magic - we could have something like this
```js
const factoryParams = {
  searchProducts: async (params) => {
    try {
     const product =  await getProductsFromApi(params);
    } catch (err) {
      if (err === "sth") const error = 'friendly error"
    }
    return { products, error }
  }
```

packages/core/rfcs/errors-habdling-in-composables.md
```diff
@@ -0,0 +1,63 @@
...
+
+Factory
+```js
+const useProductFactory = (factoryParams) => {
+  const useProduct = () => {
+    const products = ref([]);
+    const errors = ref({};)
+
+
+    const search = async (params) => {
```
patzick, line 18:
```suggestion
    const search = async (params) => {
      errors.searchProducts = null
```

packages/core/rfcs/errors-habdling-in-composables.md
```diff
@@ -0,0 +1,63 @@
...
+      try {
+        products.value = factoryParams.searchProducts(params);
+      } catch (err) {
+        errors.searchProducts = factoryParams.convertError(err);
```
lukeromanowicz, line 22: params should also be passed to the converError in order to supply additional information about context of this method used.

If we execute one method multiple times we will lose all errors except the one from last execution

PR #342, opened 2020-04-03
[RFC]  Internal fields

andrzejewsky:

### DIFF
--- a/packages/core/rfcs/setting-composable-internal-fields.md
+++ b/packages/core/rfcs/setting-composable-internal-fields.md
@@ -1,0 +1,55 @@
+# Setting composable internal fields
+
+## Motivation
+
+Sometimes when we create a new composable (using factory) we need to override some fields that are inside.
+For example, when you use `useProduct` you want to reload `cart` property from the `useCart` composable.
+To do this our factories can return setter function instead of just composable instance:
+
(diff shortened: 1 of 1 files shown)


filrak (approved):
I like that but we shouldn't provide that for every comparable, only the ones that have a shared state (so cart, user, locale (?) )

packages/core/rfcs/setting-composable-internal-fields.md
```diff
@@ -0,0 +1,55 @@
...
+```js
+import { useCartFactory } from '@vue-storefront/factories';
+
+export default useCartFactory({ ... })
+```
+
+now usage (example with user):
+
+```js
+import { setCart } from './useCart';
```
patzick, line 38: would need to be like this
```suggestion
import { useCart } from 'composablesPackage';

const {setCart} = useCart()
```

patzick (approved):
gives more flexibility for sure, but as @filrak mentioned - only makes sense for compostables with shared state :)

packages/core/rfcs/setting-composable-internal-fields.md
filrak: no, that would be an internal api from factory, Andrzej was right

PR #344, opened 2020-04-05
rfc: products filters

Qrzy:
[See rendered](https://github.com/DivanteLtd/next/blob/rfc-products-filters/packages/core/rfcs/20200417-products-filters.md)


filrak:
My suggestion would be to have `availableFilters` and `selectedFilters` (both computed) that are setted after invoking `search`

filrak:
It just came to my mind - why we need `selectedFilters` when its the same value as one passed to `search({filters})`

packages/core/rfcs/20200401-products-filters.md
```diff
@@ -0,0 +1,81 @@
...
+  selectedFilters: Filter<FILTER_BASE>[];
+}
+```
+
+### Filter interface
+
+We need a unified interface for filters, like:
+
+```typescript
+// FILTER_BASE is an enum with available filter bases, like price, attribute, status, quantity, availability, sale percentage etc.
```
filrak, line 48: whats the difference between `filterBase` and `type?

packages/core/rfcs/20200401-products-filters.md
```diff
@@ -0,0 +1,81 @@
...
+// this will be handled per integration, but basic agnostic bases are added
+interface Filter<FILTER_BASE> {
+  name: string;
+  type: FilterType;
+  filterBase: AgnosticFilterBase | FILTER_BASE;
+  options: FilterOption[]
+}
+
+enum FilterType {
+  RANGE = 'range', // eg. price
```
filrak: i dont think every platform is providing something like that, its helpful but maybe too ui-specific? if its not available everywhere imho it shouldn't be there. In theory in integration people could map every possible filter to such value but that would be a massive work far exceeding the benefits. Imho it should be decided on the UI as it was in vsf 1

packages/core/rfcs/20200401-products-filters.md
```diff
@@ -0,0 +1,81 @@
...
+  ...
+}
+
+interface FilterOption {
+  label: string;
+  value: string | number | [number, number]; // tuple for multiple min-max ranges
+  count?: number;
+}
+
+enum AgnosticFilterBase {
```
filrak: you cant assume things like that

packages/core/rfcs/20200401-products-filters.md
```diff
@@ -0,0 +1,81 @@
...
+  RANGE = 'range', // eg. price
+  SINGLE = 'single', // eg. availability status
+  MULTI = 'multi', // eg. predefined price ranges or sizes
+  ...
+}
+
+interface FilterOption {
+  label: string;
+  value: string | number | [number, number]; // tuple for multiple min-max ranges
+  count?: number;
```
filrak: count of what?

filrak (commented):
Also this will mean that we should remove `getAttributes(ProductVariant[])` as its the same functionality?

packages/core/rfcs/20200401-products-filters.md
Qrzy: I don't think it is ui-specific. Such filter type not only allows ui to choose proper component or its settings, but also provides a hint for underlying logic - how should we map filter options to something that api client will handle in request. 
So `range` will determine `min < x < max`, `single` will only allow single option to be selected, `multi` will allow multiple options to be selected and apply logical alternative on them.

packages/core/rfcs/20200401-products-filters.md
Qrzy: Count of items meeting the filter option criteria. To be delivered by search function.

packages/core/rfcs/20200401-products-filters.md
Qrzy: Could we discuss it further? Every product has price and some attributes, so I think these two are safe choice for such agnostic enum. It could be fully on intergration side, sure, but why can't it be here?

packages/core/rfcs/20200401-products-filters.md
filrak: I get that but you're not addressing my concern :), It is still additional work for the integrators so the question is - does it provide enough value to justify this additional work? In my opinion it's not as it could lead to significant amount of additional work in some cases because you can have multiple filterable fields and all of them are depending on platform and sometimes on settings from the admin panel. As a rule of thumb we shouldn't add anything that is not coming from the platform itself. We aim to lower the efforts needed to make an integration, not the opposite. I would see something like that more as an API extension than something mandatory for the integration

Qrzy:
> It just came to my mind - why we need `selectedFilters` when its the same value as one passed to `search({filters})`

Well, actually we don't need them, you are right. :) 

> Also this will mean that we should remove `getAttributes(ProductVariant[])` as its the same functionality?

Is it the same?

packages/core/rfcs/20200401-products-filters.md
filrak: Also not every platform is providing such stuff, in many cases you don't even have searchable properties predefined in a platform, only a list of attributes so in that case if anything a helper  like `getAttributesCount(product[])` or sth like that

packages/core/rfcs/20200401-products-filters.md
filrak: mostly because it can be dynamic, for example in commercetools you're specifying which fields are searchable in admin panel

packages/core/rfcs/20200401-products-filters.md
```diff
@@ -0,0 +1,57 @@
...
+
+Since there are several factors that the filters depend on (products type, category, available products, their attributes) and the set of filters might be dynamic, this feature should introduce new field in the `useProduct` composable, `availableFilters`, like in the example below.
+
+**New interface**:
+
+```ts
+export interface UseProduct<PRODUCT, FILTER_OPTION> {
+  products: ComputedProperty<PRODUCT[]>;
+  totalProducts: ComputedProperty<number>;
+  availableFilters: ComputedProperty<Filter<FILTER_OPTION>[]>;
```
lukeromanowicz: IMHO this field should be optional for platforms that don't send available filters.

packages/core/rfcs/20200401-products-filters.md
```diff
@@ -0,0 +1,57 @@
...
+  search: (params: {
+    perPage?: number;
+    page?: number;
+    sort?: any;
```
lukeromanowicz: I think we have to propose a more strict type for that, otherwise, we will not be able to implement it in the core theme

Qrzy: As I remeber from some discussions, the conclusion was that it's nearly impossible to create fully platform-agnostic type for filter options, thus I propose here a solution that assumes filter options are platform-specific and listing filters with their options will be handled in integration.

filrak:
1. Load filters in the `productSearch`
2. `productsSearch` should also return `availableFilters`
3. `availabelFilters` should have a generic type
4. we are getting rid of proposed types

packages/core/rfcs/20200417-products-filters.md
```diff
@@ -0,0 +1,47 @@
...
+
+**`UseProductFactoryParams` interface**
+
+Loading filters, their possible values, counts, etc. will require an integration-specific logic that will be part of the search function, so the factory params interface needs a new interface for search function result:
+
+```ts
+export interface ProductsSearchResult<PRODUCT, PRODUCT_FILTER> {
+  data: PRODUCT[],
+  total: number,
+  availableFilters?: PRODUCT_FILTER[];
```
filrak: whole type should be generic (`PRODUCT_FILTERS`)

Qrzy: done

andrzejewsky (approved):

PR #348, opened 2020-04-08
chore(commercetools): UseLocale, rollback

andrzejewsky:


packages/core/theme-module/theme/pages/Category.vue
```diff
@@ -163,7 +163,7 @@
             :score-rating="3"
             :isOnWishlist="false"
             @click:wishlist="toggleWishlist(i)"
-            :link="`/p/${productGetters.getSlug(product)}`"
+            :link="`/p/${productGetters.getId(product)}/${productGetters.getSlug(product)}`"
```
Qrzy, line 166: I think the same should be applied in `RelatedProducts.vue`.

packages/core/theme-module/theme/pages/Home.vue
```diff
@@ -32,6 +32,8 @@
       image="/homepage/newsletter.jpg"
       class="call-to-action"
     />
+          test: {{ $t('welcome') }}
```
Qrzy: Why so?

Qrzy (changes requested):

packages/core/theme-module/theme/pages/Home.vue
andrzejewsky: because WHO said we have to test a lot, so...

Qrzy (approved):

PR #360, opened 2020-04-08
feat: implemented user  order details

### LINKED ISSUE #197: I want to see the details of historical order So I can deeply analyze it against the one that I received in real life
**Context**
In https://github.com/DivanteLtd/next/issues/190 task we prepared the grid with order history. Now we want to let the user be able to open a single specific order to see more details of it.

**Acceptance Criteria**
* In My Orders (grid view) I can see the possibility to open details of single specific order.
* When I click on the button then I am redirected to the page with order details of single order.
* I can see its billing information.
* I can see its shipping information.
* I can see its status.
* I can see its number.
* I can see order positions with names, quantities, prices, values, discounts.
* I can see summary of the order values (net, vat, gross).

defudef:

### DIFF
--- a/packages/commercetools/composables/src/getters/orderGetters.ts
+++ b/packages/commercetools/composables/src/getters/orderGetters.ts
@@ -1,5 +1,5 @@
 import { UserOrderGetters, AgnosticOrderStatus } from '@vue-storefront/interfaces';
-import { Order, OrderState, LineItem, Money, Address, TaxPortion } from './../types/GraphQL';
+import { Order, OrderState, LineItem, Money, Address } from './../types/GraphQL';
 
 export const getOrderDate = (order: Order): string => order?.createdAt || '';
 
@@ -22,36 +22,19 @@ export const getOrderItems = (order: Order): LineItem[] => order?.lineItems || [
 
 export const getOrderPrice = (order: Order): number | null => getPrice(order?.totalPrice);
 
-export const getOrderNetValue = (order: Order): number | null => getPrice(order?.taxedPrice?.totalNet);
-
-export const getOrderGrossValue = (order: Order): number | null => getPrice(order?.taxedPrice?.totalGross);
-
-export const getOrderTaxes = (order: Order): TaxPortion[] => order?.taxedPrice?.taxPortions || [];
-
-export const getOrderTaxName = (taxPortion: TaxPortion): string => taxPortion.name || '';
-
-export const getOrderTaxValue = (taxPortion: TaxPortion): number | null => getPrice(taxPortion.amount);
-
-export const getOrderTaxRate = (taxPortion: TaxPortion): number => (taxPortion.rate || 0) * 100;
-
-export const getOrderBillingAddress = (order: Order): Address | null => order?.billingAddress || null;
-
-export const getOrderShippingAddress = (order: Order): Address | null => order?.shippingAddress || null;
-
-type KeyValueAddress = { property: string; value: string }[];
-
-const transformAddressToArray = (address: Address | object): KeyValueAddress => Object.entries(address)
+const transformAddressToString = (address: Address | object): string => Object.entries(address)
   .filter(([property, value]) => value !== null && !['id', '__typename'].includes(property))
-  .map(([property, value]) => ({
-    property: String(property)
-      .replace(/([A-Z])/g, ' $1')
-      .replace(/^./, (str) => str.toUpperCase()),
-    value
-  }));
+  .map(([property, value]) => (
+    '<div class="order-details__addresses__item__field">' +
+    `<span class="property">${property.replace(/([A-Z])/g, ' $1').replace(/^./, (str) => str.toUpperCase())}</span>` +
+    `<span class="value">${value}</span>` +
+    '</div>'
+  ))
+  .join('');
 
-export const getOrderBillingAddressValues = (order: Order): KeyValueAddress => transformAddressToArray(getOrderBillingAddress(order) || {});
+export const getOrderBillingAddress = (order: Order): string => transformAddressToString(order?.billingAddress || {});
 
-export const getOrderShippingAddressValues = (order: Order): KeyValueAddress => transformAddressToArray(getOrderShippingAddress(order) || {});
+export const getOrderShippingAddress = (order: Order): string => transformAddressToString(order?.shippingAddress || {});
 
 const orderGetters: UserOrderGetters<Order> = {
   getDate: getOrderDate,
@@ -60,16 +43,8 @@ const orderGetters: UserOrderGetters<Order> = {
   getStatus: getOrderStatus,
   getPrice: getOrderPrice,
   getItems: getOrderItems,
-  getNetValue: getOrderNetValue,
-  getGrossValue: getOrderGrossValue,
-  getTaxes: getOrderTaxes,
-  getTaxName: getOrderTaxName,
-  getTaxValue: getOrderTaxValue,
-  getTaxRate: getOrderTaxRate,
   getBillingAddress: getOrderBillingAddress,
-  getShippingAddress: getOrderShippingAddress,
-  getBillingAddressValues: getOrderBillingAddressValues,
-  getShippingAddressValues: getOrderShippingAddressValues
+  getShippingAddress: getOrderShippingAddress
 };
 
 export default orderGetters;
--- a/packages/commercetools/composables/src/getters/orderGetters.ts
+++ b/packages/commercetools/composables/src/getters/orderGetters.ts
@@ -22,16 +22,13 @@ export const getOrderItems = (order: Order): LineItem[] => order?.lineItems || [
 
 export const getOrderPrice = (order: Order): number | null => getPrice(order?.totalPrice);
 
-const transformAddressToString = (address: Address | object): string => Object.entries(address)
-  .filter(([property, value]) => value !== null && !['id', '__typename'].includes(property))
-  .map(([property, value]) => (
-    `${property.replace(/([A-Z])/g, ' $1').replace(/^./, (str) => str.toUpperCase())}: ${value}`
-  ))
-  .join(', ');
+const transformAddressToString = (address: Address): string => (
+  `${address.country}, ${address.postalCode}, ${address.city}, ${address.streetName}, ${address.streetNumber}`
+);
 
-export const getOrderBillingAddress = (order: Order): string => transformAddressToString(order?.billingAddress || {});
+export const getOrderBillingAddress = (order: Order): string => transformAddressToString(order?.billingAddress || {} as Address);
 
-export const getOrderShippingAddress = (order: Order): string => transformAddressToString(order?.shippingAddress || {});
+export const getOrderShippingAddress = (order: Order): string => transformAddressToString(order?.shippingAddress || {} as Address);
 
 const orderGetters: UserOrderGetters<Order> = {
   getDate: getOrderDate,
--- a/packages/commercetools/composables/src/getters/orderGetters.ts
+++ b/packages/commercetools/composables/src/getters/orderGetters.ts
@@ -5,8 +5,6 @@ export const getOrderDate = (order: Order): string => order?.createdAt || '';
 
 export const getOrderId = (order: Order): string => order?.id || '';
 
-export const getOrderNumber = (order: Order): string => order?.orderNumber || getOrderId(order);
-
 const orderStatusMap = {
   [OrderState.Open]: AgnosticOrderStatus.Open,
   [OrderState.Confirmed]: AgnosticOrderStatus.Confirmed,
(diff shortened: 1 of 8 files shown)


packages/core/theme-module/theme/pages/MyAccount.vue
```diff
@@ -37,18 +37,24 @@
         <SfContentPage title="My reviews">
           <MyReviews />
         </SfContentPage>
+        <SfContentPage title="Order details">
+          <OrderDetails v-if="entityId" :orderId="entityId" />
```
filrak: that logic should be a part of `OrderDetails` component, not `MyAccount`

packages/core/theme-module/theme/pages/MyAccount/OrderDetails.vue
```diff
@@ -0,0 +1,140 @@
+<template>
+  <section class="order-details">
+    <h2>Order #{{orderGetters.getNumber()}}</h2>
```
filrak:
```suggestion
    <h2>Order #{{orderGetters.getNumber(order)}}</h2>
```

filrak (changes requested):

filrak:
also coverage decreased

packages/commercetools/composables/src/getters/orderGetters.ts
```diff
@@ -1,26 +1,69 @@
...
+export const getOrderNumber = (order: Order): string => order?.orderNumber || getOrderId(order);
+
 const orderStatusMap = {
   [OrderState.Open]: AgnosticOrderStatus.Open,
   [OrderState.Confirmed]: AgnosticOrderStatus.Confirmed,
   [OrderState.Complete]: AgnosticOrderStatus.Complete,
   [OrderState.Cancelled]: AgnosticOrderStatus.Cancelled
 };
 
+const getPrice = (money: Money): number | null => money && money.centAmount ? money.centAmount / 100 : null;
```
Qrzy:
```suggestion
const getPrice = (money: Money): number | null => money?.centAmount  / 100 || null;
```

packages/commercetools/composables/src/getters/orderGetters.ts
```diff
@@ -1,26 +1,69 @@
 import { UserOrderGetters, AgnosticOrderStatus } from '@vue-storefront/interfaces';
-import { Order, OrderState } from './../types/GraphQL';
+import { Order, OrderState, LineItem, Money, Address } from './../types/GraphQL';
 
 export const getOrderDate = (order: Order): string => order?.createdAt || '';
 
 export const getOrderId = (order: Order): string => order?.id || '';
 
+export const getOrderNumber = (order: Order): string => order?.orderNumber || getOrderId(order);
```
Qrzy: Should such fallback be hidden like this under the getter?

packages/commercetools/composables/src/getters/orderGetters.ts
```diff
@@ -1,26 +1,69 @@
...
+
+export const getOrderPrice = (order: Order): number | null => getPrice(order?.totalPrice);
+
+export const getOrderNetValue = (order: Order): number | null => getPrice(order?.taxedPrice?.totalNet);
+
+export const getOrderGrossValue = (order: Order): number | null => getPrice(order?.taxedPrice?.totalGross);
+
+export const getOrderTaxValue = (order: Order): number | null => getPrice(order?.taxedPrice?.taxPortions[0]?.amount);
+
+export const getOrderTaxRate = (order: Order): number => (order?.taxedPrice?.taxPortions[0]?.rate || 0) * 100;
```
Qrzy: What if items in the order have different tax rates? I think this helper should return an array.

Qrzy (changes requested):

packages/commercetools/composables/src/getters/orderGetters.ts
defudef: I think it should. If the API doesn't create an order number, then we should use the ID of the entity. It's more agnostic I think

andrzejewsky (approved):

packages/commercetools/composables/src/getters/orderGetters.ts
```diff
@@ -1,26 +1,46 @@
-import { UserOrderGetters, AgnosticOrderStatus } from '@vue-storefront/interfaces';
-import { Order, OrderState } from './../types/GraphQL';
+import { UserOrderGetters, AgnosticOrderStatus, AgnosticPrice } from '@vue-storefront/interfaces';
+import { Order, OrderState, LineItem, Money, Address } from './../types/GraphQL';
 
 export const getOrderDate = (order: Order): string => order?.createdAt || '';
 
 export const getOrderId = (order: Order): string => order?.id || '';
 
+export const getOrderNumber = (order: Order): string => order?.orderNumber || getOrderId(order);
```
filrak: how that is different from ID? Did you made sure that this field is not specific to commercetools?

packages/commercetools/composables/src/getters/orderGetters.ts
```diff
@@ -1,26 +1,46 @@
...
+});
+
 export const getOrderStatus = (order: Order): AgnosticOrderStatus | '' => order?.orderState ? orderStatusMap[order.orderState] : '';
 
-export const getOrderPrice = (order: Order): number | null => order ? order.totalPrice.centAmount / 100 : null;
+export const getOrderItems = (order: Order): LineItem[] => order?.lineItems || [];
+
+export const getOrderPrice = (order: Order): AgnosticPrice => getPrice(order?.totalPrice);
+
+const transformAddressToString = (address: Address): string => (
```
filrak, line 26: shouldn't that be a set of helpers OR just a non-agnostic object?

@andrzejewsky we've just discussed that

packages/core/theme-module/theme/pages/MyAccount.vue
filrak: if entityID is taken from URL it should be a part of `<OrderDetails>` component imho

packages/core/theme-module/theme/pages/MyAccount/OrderDetails.vue
```diff
@@ -0,0 +1,144 @@
...
+  props: ['orderId'],
+  setup({ orderId }) {
+    const { searchOrders, orders } = useUserOrders();
+
+    onSSR(async () => {
+      await searchOrders({ id: orderId });
+    });
+
+    return {
+      cartGetters,
```
filrak, line 63: I'm not entirely sure about that so its up for discussion but maybe we should just have item getters in `orderGetters` are we sure that in every platform item from order is the same one as cart item? Also from the DX point of view It doesn't seem intuitive, for every other comparable you have a single corresponding getter
@defudef @andrzejewsky whats your opinion

packages/core/theme-module/theme/pages/MyAccount/OrderDetails.vue
```diff
@@ -0,0 +1,144 @@
...
+
+    &__field {
+      display: flex;
+    }
+  }
+
+}
+</style>
+
+<style lang="scss">
```
filrak, line 113: why do we need to have that as a shared style

packages/core/theme-module/theme/pages/MyAccount/OrderHistory.vue
```diff
@@ -30,7 +30,9 @@
           </SfTableData>
           <SfTableData class="orders__view">
             <SfButton class="sf-button--text color-secondary mobile-only">Download</SfButton>
-            <SfButton class="sf-button--text color-secondary desktop-only">VIEW</SfButton>
+            <nuxt-link :to="{ path: `/my-account/order-details/${orderGetters.getId(order)}`}">
```
filrak, line 33: use localized link ;)

filrak (changes requested):

packages/core/theme-module/theme/pages/MyAccount/OrderDetails.vue
defudef: I think we need in this case orderItemGetters with OrderItem component. Unfortunately, there is no component in SFUI for that case

packages/core/theme-module/theme/pages/MyAccount/OrderDetails.vue
defudef: scoped style doesn't affect on SfCollectedProduct elements

packages/commercetools/composables/src/getters/orderGetters.ts
defudef: It's temporary actually (just to keep DRY)

packages/core/theme-module/theme/pages/MyAccount.vue
defudef: I think it's better to provide it through props as it gives better control, but we can discuss about it

packages/commercetools/composables/src/getters/orderGetters.ts
defudef: A lot of ecommerce platforms have the ID and Order Number separately.

packages/core/theme-module/theme/pages/MyAccount/OrderHistory.vue
defudef: Can you explain? I haven't seen any localized url in the project

packages/commercetools/composables/src/getters/orderGetters.ts
filrak: if something is temporarly add an issue and todo here with issue linked

packages/core/theme-module/theme/pages/MyAccount.vue
filrak: what do you mean by better control?

packages/core/theme-module/theme/pages/MyAccount/OrderDetails.vue
filrak: I'm not sure I get it, why you want to pair component and getters?

defudef: The problem is with a component that renders bought items:
```
SfCollectedProduct
        v-for="product in orderGetters.getItems(order)"
        :key="cartGetters.getItemSku(product)"
        :image="cartGetters.getItemImage(product)"
        :title="cartGetters.getItemName(product)"
        :regular-price="'$' + cartGetters.getItemPrice(product).regular"
        :qty="cartGetters.getItemQty(product)"
        class="ordered-product"
      />
```
the orderGetters doesn't have methods that returns: SKU, image, or quantity

packages/core/theme-module/theme/pages/MyAccount.vue
defudef: By providing entityId as a prop you can get it on a way you define (or URL, or cookie, or cache). Simply: it's open for extensions. If I'll define entityId directly in OrderDetails I'm forced to get entityId from URL only.

packages/commercetools/composables/src/getters/orderGetters.ts
andrzejewsky:
```
export const getOrderId = (order: Order): string => order?.id || order?.orderNumber;

```

filrak: does a lot mean all of them? keep in mind that helpers are for the UI, are both needed in the UI? Imho @andrzejewsky proposed right solution

packages/core/theme-module/theme/pages/MyAccount.vue
filrak: no, you can still change it in the exactly same way you're changing `entityId` prop

filrak: Imho what I want to discuss here is do we pass always via props or always use composable inside for all the pages, In checkout its separate (composable per page)

defudef: as we agreed - I removed that prop

packages/core/theme-module/theme/pages/MyAccount/OrderDetails.vue
defudef: RFC proposed here: https://github.com/DivanteLtd/next/issues/363

filrak (approved):

packages/core/interfaces/src/index.ts
```diff
@@ -207,11 +207,14 @@ export interface CheckoutGetters<SHIPPING_METHOD> {
...
-export interface UserOrderGetters<ORDER> {
+export interface UserOrderGetters<ORDER, ORDER_ITEM> {
   getDate: (order: ORDER) => string;
   getId: (order: ORDER) => string;
   getStatus: (order: ORDER) => string;
-  getPrice: (order: ORDER) => number;
+  getPrice: (order: ORDER) => AgnosticPrice;
+  getItems: (order: ORDER) => ORDER_ITEM[];
+  getBillingAddress: (address: ORDER) => string;
+  getShippingAddress: (address: ORDER) => string;
```
patzick, line 217: would change that for implementation to decide about address structure
```suggestion
  getBillingAddress: (address: ORDER) => ADDRESS;
  getShippingAddress: (address: ORDER) => ADDRESS;
```

patzick (changes requested):
One thing from me, Address should be an implementation interface here :)

filrak (changes requested):
Get rid of additional getters in core

packages/core/interfaces/src/index.ts
filrak: remove and create RFC, ideally before 4 m so we can discuss it on decision making window @defudef

PR #372, opened 2020-04-10
[RFC] price type proposal

andrzejewsky:


andrzejewsky:
```ts
interface AgnosticPrice  {
  regular: number
  special?: number;
}
```

- regular and price is taxed or not when the getter is called (based on country)
- introduce filter  function that will format given currency

PR #375, opened 2020-04-10
refactor: template building overhaul

lukeromanowicz:
feat: replace lodash template with ejs
feat: rebuild single files instead of all files in dev mode
feat: add includes in theme
fix: race condition between templates and overwrite components


filrak (changes requested):
Changes as we discussed - use nuxt module options

Qrzy (approved):

andrzejewsky (approved):

packages/core/theme-module/ensureDirectoryExists.js
```diff
@@ -0,0 +1,11 @@
+const fs = require('fs');
```
filrak: maybe lets put all those files into a `scripts` folder. It slowly becomes messy

packages/core/theme-module/theme/pages/Category.vue
```diff
@@ -375,15 +376,13 @@ export default {
...
-    const lastSlug = Object.keys(params).reduce((prev, curr) => params[curr] || prev, params.slug_1);
-
     const { categories, search, loading } = useCategory('categories');
     const { products: categoryProducts, totalProducts, search: productsSearch, loading: productsLoading } = useProduct('categoryProducts');
     const currentPage = ref(parseInt(query.page, 10) || 1);
     const itemsPerPage = ref(parseInt(query.items, 10) || perPageOptions[0]);
 
     onSSR(async () => {
-      await search({ slug: lastSlug });
+      await search(getCategorySearchParameters(context));
```
filrak, line 385: I reminded you that it should be inlined. `includes` dir should exist only in `theme-module`

filrak (changes requested):

packages/core/theme-module/theme/pages/Category.vue
lukeromanowicz: We agreed that in the end, we will move this one to an external function

packages/core/theme-module/ensureDirectoryExists.js
lukeromanowicz: fixed

PR #377, opened 2020-04-13
feat: improved boilerplate package

thampe:

Hi,

I wanted to take a look at vuestorefront next. Initially I wanted to start a new api-client implementation, but the boilerplate package did not seem to work. So I updated the boilerplate package. At first it seemed like an easy/tiny fix to make it work, but there were some things missing in the boilerplate package and now I added them. Now the boilerplate package at least runs without significant errors with `yarn dev:bp`.

Could someone take a look at it? I am by no means an expert in this field, I mostly do php/magento stuff. 

Thanks and greetings
\- Thomas


rmakara:
Thank you @thampe :) we will take a look at that.

packages/boilerplate/api-client/src/api/getCategory/index.ts
```diff
@@ -0,0 +1,5 @@
+import {Category} from '../../types';
+
+export default async function (): Promise<Category[]> {
```
filrak, line 4: Have you seen [this](https://vsf-next-docs.netlify.com/integration/api-client.html) doc? Would be cool to have an example including overriding mechanism.

packages/boilerplate/api-client/src/index.ts
```diff
@@ -1,10 +1,16 @@
 import getProduct from './api/getProduct';
+import getCategory from './api/getCategory';
 import { apiClientFactory } from '@vue-storefront/factories';
 
-const { setup, override } = apiClientFactory<any, any>();
+const { setup, override } = apiClientFactory<any, any>({
+  defaultSettings: {},
+  // eslint-disable-next-line @typescript-eslint/no-unused-vars
+  onSetup: (config: any) => {}
+});
```
filrak, line 11: how about providing settings for api client methods as well?
I mean something like

```
const settings = getSettings()
```

packages/boilerplate/theme/nuxt.config.js
```diff
@@ -8,25 +13,37 @@ export default {
...
+  ],
+  buildModules: [
+    // to core
+    '@nuxt/typescript-build',
     ['@vue-storefront/nuxt', {
       coreDevelopment: true,
       useRawSource: {
         dev: [
-          '@vue-storefront/boilerplate-composables'
+          '@vue-storefront/boilerplate-composables',
```
filrak: the new convention for composables package naming is without `composables` part (eg. `@vue-storefront/commercetools`)

packages/boilerplate/theme/plugins/boilerplate-config.js
```diff
@@ -0,0 +1,32 @@
+export const config = {
+  api: {},
+  locale: 'en',
```
filrak: All those parts are integration-specific. For example - not every integration will use cookies. I think its best to keep the boilerplate config completely clean 9and also get rid of cookie universal pckg).

packages/boilerplate/theme/plugins/boilerplate.js
```diff
@@ -1,4 +1,4 @@
 import { setup } from '@vue-storefront/boilerplate-api';
+import { config } from './boilerplate-config';
```
filrak: How about keeping it in one file (config in `boilerplate.js`) or making a folder to group those files? It doesn't seem right for me to have a file that is not a plugin in `plugins` dir (i know we have it in CT integration - I'm also thinking about a better way to structure this)

filrak (commented):
Hey! 

This is absolutely amazing piece of work and will help us a lot in upcoming integrations! Thank you so much!

image

I left few suggestions on how we can improve the boilerplate and give people even a better base. Please let me know what you think about them :)

Also please don't get scared of conflicts. We just changed one path so it should be straightforward to fix that ( `utils` `interfaces` and `factories` were merged to `@vue-storefront/core`(

Regarding changes - we can group them into a new issue to merge the PR faster so people can already apply new APIs to boilerplate as well in their PRs, wdyt?

packages/boilerplate/api-client/src/api/getCategory/index.ts
thampe: No Problem. I have added the overrides to the example api implementations.

packages/boilerplate/api-client/src/index.ts
thampe: 👍

packages/boilerplate/theme/nuxt.config.js
thampe: 👍

packages/boilerplate/theme/plugins/boilerplate-config.js
thampe: Ok. I have removed the config entirely.

packages/boilerplate/theme/plugins/boilerplate.js
thampe: 👍

PR #380, opened 2020-04-14
feat(commercetools): Display prices

andrzejewsky:


packages/commercetools/composables/src/getters/_utils.ts
```diff
@@ -57,3 +58,28 @@ export const getVariantByAttributes = (products: ProductVariant[] | Readonly<Pro
   });
 };
 
+export const createPrice = (product: ProductVariant | LineItem): AgnosticPrice => {
+  const getPrice = (price: ProductPrice | DiscountedProductPriceValue) => price ? price.value.centAmount / 100 : null;
+  const getDiscount = (product: ProductVariant | LineItem) => product.price?.discounted;
```
Qrzy: Why to define those helper functions on every function call when they don't use any variables from the function scope?
It would be more efficient to move them outside:

```suggestion
const getPrice = (price: ProductPrice | DiscountedProductPriceValue) => price ? price.value.centAmount / 100 : null;
const getDiscount = (product: ProductVariant | LineItem) => product.price?.discounted;

export const createPrice = (product: ProductVariant | LineItem): AgnosticPrice => {
```

Qrzy (approved):

filrak (changes requested):
IMHO if its a part of `product GETTERS it should be called `getFormattedPrice`

filrak (approved):

PR #381, opened 2020-04-14
ref: useUserAddressFactory

defudef:


packages/core/rfcs/use-user-address-factory.md
```diff
@@ -0,0 +1,128 @@
...
+
+## Motivation
+
+We'd like to separate a logic for the user addresses according to single responsibility principle. It's also more agnostic approach.
+
+## Factory interface
+
+```TS
+// factory schema
+export interface UseUserAddressFactory<ADDRESS, SEARCH_PARAMS> {
```
filrak: Do you know that there is already an interface in the core? You should mention that in RFC and tell why you want to add certain properties to avoid discussions on unnecessary stuff.

packages/core/rfcs/use-user-address-factory.md
```diff
@@ -0,0 +1,128 @@
...
+  addresses: ComputedProperty<ADDRESS[]>;
+  totalAddresses: ComputedProperty<number>;
+  searchAddresses: (params?: SEARCH_PARAMS) => Promise<void>;
+  addShippingAddress: (address: ADDRESS) => Promise<void>;
+  addBillingAddress: (address: ADDRESS) => Promise<void>;
+  getShippingAddresses: () => ADDRESS[];
```
filrak: that shouldnt be a getX function (or if it would remain like that it should be a getter but probably that wasn;t your intention) since it collides with getters naming.

`billingAddresses` seems like a better way of naming it BUT then you have a duplication of data (`addressess` will contain a superset of `billingAddresses`). Can we have more than 1 assigned billing addr/shipping adds since you used a plural form here?

If we don't shouldn't that be just a computed property? is it fetched separately?

Another option would be to just have a getter `getBillingAddresses(address: ADDRESS[])`

packages/core/rfcs/use-user-address-factory.md
```diff
@@ -0,0 +1,128 @@
...
+  loading: ComputedProperty<boolean>;
+}
+```
+
+## Factory params
+
+```TS
+// factory params schema
+export type UseUserAddressFactoryParams<ADDRESS, SEARCH_PARAMS> = {
+  searchAddresses: (params?: SEARCH_PARAMS) => Promise<ADDRESS[]>;
```
filrak: according to our convention `loadAddresses` or `addressesSearch` if parametrized

packages/core/rfcs/use-user-address-factory.md
```diff
@@ -0,0 +1,128 @@
...
+  addShippingAddress: (address: ADDRESS) => Promise<ADDRESS[]>;
+  addBillingAddress: (address: ADDRESS) => Promise<ADDRESS[]>;
```
filrak: how we add address that is not marked as billing or shipping?

packages/core/rfcs/use-user-address-factory.md
```diff
@@ -0,0 +1,128 @@
...
+  getShippingAddresses: () => ADDRESS[];
+  getBillingAddresses: () => ADDRESS[];
+  deleteAddress: (address: ADDRESS) => Promise<ADDRESS[]>;
+  updateAddress: (address: ADDRESS) => Promise<ADDRESS[]>;
+}
+```
+
+## User address for Commerce Tools purposes
```
filrak: I think its already solved here but it's integration-specific stuff therefore shouldn't be a part of the RFC https://github.com/DivanteLtd/next/pull/342

packages/core/rfcs/use-user-address-factory.md
defudef: Will we have that case? If so then we need additional methods:
```TS
type addAddress = (address: ADDRESS) => Promise<ADDRESS[]>;
type assignAddressToBilling = (address: ADDRESS) => Promise<ADDRESS>;
type assignAddressToShipping = (address: ADDRESS) => Promise<ADDRESS>;
```
Or something like that

packages/core/rfcs/use-user-address-factory.md
defudef: Computed value makes sens for me

packages/core/rfcs/use-user-address-factory.md
```diff
@@ -0,0 +1,153 @@
...
+
+### 2nd approach (shorter):
+```TS
+export interface UseUserAddressFactory<ADDRESS, ADDRESS_TYPE = 'billing' | 'shipping', SEARCH_PARAMS> {
+  addresses: ComputedProperty<ADDRESS[]>;
+  shippingAddresses: ComputedProperty<ADDRESS[]>;
+  billingAddresses: ComputedProperty<ADDRESS[]>;
+  totalAddresses: ComputedProperty<number>;
+  loadAddresses: (params?: SEARCH_PARAMS) => Promise<void>;
+  addAddress: (address: ADDRESS, type?: ADDRESS_TYPE) => Promise<void>;
```
patzick:
```suggestion
  addAddress: (address: ADDRESS, options?: unknown) => Promise<void>;
```

packages/core/rfcs/use-user-address-factory.md
```diff
@@ -0,0 +1,112 @@
...
+};
+```
+
+### Factory content
+```TS
+const useUserAddressFactory = () => {
+  const addresses: ref([]);
+  const loading: ref(false);
+
+  const addAddress: async (address) => {
```
defudef: Sorry for shorthand. It should be obviously be like this:
```TS
const useUserAddressFactory = () => {
  const addresses = ref([]);
  const loading = ref(false);

  return function useUserAddress(): UseUserAddress<ADDRESS, SEARCH_PARAMS, ADDRESS_OPTIONS> {
    const addAddress: async (address) => {
      loading.value = true;
      addresses.value = await factoryParams.addAddress(address);
      loading.value = false;
    }

    return {
      addresses: computed(() => addresses.value),
      loading: computed(() => loading.value),
      // ...other values here
    }
  }
};
```

packages/core/rfcs/use-user-address-factory.md
```diff
@@ -0,0 +1,112 @@
+# UseUserAddressFactory with Commerce Tools composable use case
+
+## Motivation
+
+We need to extend useUserAddress type to show and modify user addresses.
+Also, some eCommerce platforms requires updating user entity with every address modification (for example: CommerceTools).
+
+## Factory interface
```
defudef: That's obviously a composable type made be a factory

packages/core/rfcs/use-user-address-factory.md
```diff
@@ -0,0 +1,120 @@
...
+// FACTORY INTERFACE
+export interface UseUser
+<
+  USER,
+  UPDATE_USER_PARAMS
+> {
+  // ...leave all the stuff as they are now
+
+  // needed to add this:
+  setUser: (newUser: USER) => void;
```
filrak, line 47: It shouldn't be a part of params, we have access to this object from the inside therefore we can just return this fn

filrak: Please take a look at @andrzejewsky RFC

filrak (changes requested):

packages/core/rfcs/use-user-address-factory.md
```diff
@@ -0,0 +1,120 @@
...
+## Motivation
+
+We need to extend useUserAddress type to show and modify user addresses.
+Also, some eCommerce platforms requires updating user entity with every address modification (for example: CommerceTools).
+
+## Composable interface
+```TS
+export interface UseUserAddress<ADDRESS, SEARCH_PARAMS, ADDRESS_OPTIONS = unknown> {
+  addresses: ComputedProperty<ADDRESS[]>;
+  shippingAddresses: ComputedProperty<ADDRESS[]>;
```
filrak, line 12: can we have multiple shipping addressess and billing addresses? Does every address have to be in a specific category?

packages/core/rfcs/use-user-address-factory.md
```diff
@@ -0,0 +1,120 @@
...
+  billingAddresses: ComputedProperty<ADDRESS[]>;
+  loadAddresses: (params?: SEARCH_PARAMS) => Promise<void>;
+  addAddress: (address: ADDRESS, options?: ADDRESS_OPTIONS) => Promise<void>;
```
filrak, line 15: what are these `options`

packages/core/rfcs/use-user-address-factory.md
```diff
@@ -0,0 +1,120 @@
...
+  updateAddress: (address: ADDRESS, options?: ADDRESS_OPTIONS) => Promise<void>;
```
patzick, line 16:
```suggestion
  saveAddress: (address: ADDRESS, options?: ADDRESS_OPTIONS) => Promise<void>;
```

patzick (changes requested):
* We'll leave only addresses, getters may filter shipping and billing addresses from it.
There should be an additional field in address, which would be considered by getters to filter addresses of different types. ( ex `__type`)
* for saving address there is only `saveAddress` method

PR #407, opened 2020-04-17
[RFC] Discount getters

andrzejewsky:

### DIFF
--- a/packages/core/rfcs/discount-getters.md
+++ b/packages/core/rfcs/discount-getters.md
@@ -1,0 +1,56 @@
+# Discount / coupons getters
+
+## Motivation
+
+There is no possibility to read discounts from the cart. We are not able to display discounted total price or even an information about applied discount code. Suggestion bellow:
+
+```ts
+export interface CartGetters<CART, CART_ITEM, DISCOUNT> {
+  // ...
+  // new ones:
+
+  getDiscount: () => DISCOUNT; // don't think if it's needed
+  getDiscountId: () => string;
+  getDiscountedTotals: () => AgnosticTotals;
+  getDiscountName: () => string;
+  getDiscountValue: () => number;
+  getDiscountType: () => string; // fixed or percentage
+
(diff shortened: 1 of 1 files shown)


packages/core/rfcs/discount-getters.md
```diff
@@ -0,0 +1,56 @@
...
+export interface CartGetters<CART, CART_ITEM, DISCOUNT> {
+  // ...
+  // new ones:
+
+  getDiscount: () => DISCOUNT; // don't think if it's needed
+  getDiscountId: () => string;
+  getDiscountedTotals: () => AgnosticTotals;
+  getDiscountName: () => string;
+  getDiscountValue: () => number;
+  getDiscountType: () => string; // fixed or percentage
```
filrak, line 17: it's not necessary

packages/core/rfcs/discount-getters.md
```diff
@@ -0,0 +1,56 @@
...
+```ts
+export interface CartGetters<CART, CART_ITEM, DISCOUNT> {
+  // ...
+  // new ones:
+
+  getDiscount: () => DISCOUNT; // don't think if it's needed
+  getDiscountId: () => string;
+  getDiscountedTotals: () => AgnosticTotals;
+  getDiscountName: () => string;
+  getDiscountValue: () => number;
```
filrak, line 16: should return string eg "-15%" "-20 złotych"

packages/core/rfcs/discount-getters.md
```diff
@@ -0,0 +1,56 @@
...
+  getDiscountType: () => string; // fixed or percentage
+
+  [getterName: string]: (element: any, options?: any) => unknown;
+}
+```
+
+Also I'm thinking about separated getters, which has even more sense as you can have multiple discounts applied to the cart.
+
+```ts
+export interface DiscountGetters {
```
filrak, line 26: Can be different between platforms

